PR #12, opened 2023-01-04
Feature/auth

### LINKED ISSUE #5: auth: support for authentication

bdmendes:
Closes #5

As discussed in meetings, for authenticating, we're going to fetch the SvelteKit server which acts as a proxy to the Spring backend - it is responsible for requesting/refreshing tokens if the request needs it, storing the tokens in the browser with `HttpOnly` cookies, for security reasons. This means that since all cookies all automatically sent by the browser, frontend developers mostly won't need to worry about auth. Also, to remove the cookies, a new `logout` endpoint was created (JS is not able to perform this action, since the cookies are `HttpOnly`).

This solution also allows for `SSR` for a more "web-like" experience.

Note that I've included some HTML pages to showcase the login/logout/profile potential. These will of course be replaced later.


bdmendes:
This is now ready for review. I've updated the description accordingly.

jamcunha (approved):
LGTM

.env.local
```diff
@@ -0,0 +1 @@
+PUBLIC_API_URL=http://localhost:8080
```
BrunoRosendo: .env.local shouldn't be in the repo. The repo should only track .env and the local overrides it

.gitignore
```diff
@@ -5,8 +5,6 @@ node_modules
 /.svelte-kit
 /package
 .env
-.env.*
-!.env.example
```
BrunoRosendo, line 9: Related to above

src/hooks.server.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  const refreshToken = json[JWT_REFRESH_KEY];
+  if (refreshToken) {
+    appendCookieHeader(serverResponse, JWT_REFRESH_KEY, refreshToken);
+  }
+
+  return serverResponse;
+}
+
+export const handle: Handle = async function ({ event, resolve }) {
```
BrunoRosendo, line 26:
```suggestion
export async function handle({ event, resolve }): Handle {
```
Any reason to not use this syntax or an arrow function? Assigning a variable like this seems odd

src/lib/auth.ts
```diff
@@ -0,0 +1,11 @@
+export const JWT_ACCESS_KEY = 'access_token';
+export const JWT_REFRESH_KEY = 'refresh_token';
```
BrunoRosendo: Maybe these should live in .env? I know sveltekit already handles .env logic
https://kit.svelte.dev/docs/modules#$env-dynamic-private

src/routes/api/[...endpoint]/proxy.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (!jwt) {
+    return _fetchAdapter(relativeUrl, method, body, new Headers(headers));
+  }
+
+  const authHeaders = new Headers(headers);
+  authHeaders.append('Authorization', `Bearer ${jwt}`);
+
+  const response = await _fetchAdapter(relativeUrl, method, body, authHeaders);
+
+  const unauthorized = response.status >= 401 && response.status <= 403;
```
Comment (BrunoRosendo) [inline src/routes/api/[...endpoint]/proxy.ts:47]: Do we have the `StatusCodes` package? It'd be nice for readability, I think this is the most common one https://www.npmjs.com/package/http-status-codes

src/routes/api/[...endpoint]/proxy.ts
```diff
@@ -0,0 +1,71 @@
+import { PUBLIC_API_URL } from '$env/static/public';
+import type { Cookies } from '@sveltejs/kit';
+import { JWT_REFRESH_KEY, JWT_ACCESS_KEY, appendCookieHeader } from '$lib/auth';
+
+async function _fetchAdapter(
```
Comment (BrunoRosendo) [inline src/routes/api/[...endpoint]/proxy.ts]: This function is used a lot, so it's nice to have it but I don't entirely understand its goal. Is it just to prepend the API URL and add the json header? If so, maybe we should have the name reference API in some way?

src/routes/api/auth/logout/+server.ts
```diff
@@ -0,0 +1,33 @@
...
+    JSON.stringify({
+      message: 'You have been logged out'
+    }),
+    {
+      status: 200
+    }
+  );
+  response.headers.append(
+    'Set-Cookie',
+    `${JWT_ACCESS_KEY}=; HttpOnly; Max-Age=0; Path=/; SameSite=Strict`
```
BrunoRosendo: What do you think of having a function to append a header that destroys cookies? The same way we have in `$lib/auth` to set cookies

src/routes/api/auth/logout/+server.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  const response = new Response(
+    JSON.stringify({
+      message: 'You have been logged out'
+    }),
+    {
+      status: 200
+    }
+  );
```
BrunoRosendo, line 24: One could think about abstracting these responses but we'll probably never do something like this again so I don't think it's worth it. Still dropping the idea though

BrunoRosendo (changes requested):
Good job! I'm still a bit confused about some details but I think this makes sense and creates a nice abstraction for fetching future requests. I left some minor suggestions and will talk to you for additional info

src/hooks.server.ts
bdmendes: Wouldn't that imply that the function returns Handle? In my implementation, Handle is the type of the function

BrunoRosendo: Ohh that makes sense, never mind then. The docs uses the `satisfies` but it should be the same

src/routes/api/auth/logout/+server.ts
bdmendes: I'll subscribe to what you just said, understand but do not think it's worth it given the use case

src/routes/api/[...endpoint]/proxy.ts
Comment (bdmendes) [reply src/routes/api/[...endpoint]/proxy.ts:47]: We do not; since this is a single use case, I would prefer to not include an extra package for now

src/routes/api/[...endpoint]/proxy.ts
Comment (bdmendes) [reply src/routes/api/[...endpoint]/proxy.ts]: Changed the name accordingly

.env
```diff
@@ -0,0 +1,4 @@
+PUBLIC_API_URL=http://localhost:8080
+PUBLIC_JWT_ACCESS_KEY=access_token
+PUBLIC_JWT_REFRESH_KEY=refresh_token
```
BrunoRosendo, line 3: Should these be public? I'm not sure myself

BrunoRosendo (approved):
I'm approving but I just wanted to clarify the public/private env variables, since it's possible to restrict some env to the server

.env
bdmendes: Yes, it's part of the API specification. The SvelteKit server does not store tokens in any way, too

PR #13, opened 2023-01-04
Implements layout

marineve17:
Closes issue #4 

- [ ] Layout file with background and colours
- [ ] Navbar in PC layout
- [ ] Navbar in mobile layout
- [ ] Footer in PC layout
- [ ] Footer in mobile layout

PR #29, opened 2023-01-09
Feature/error-page

### LINKED ISSUE #9: errors: create error page
Check out `+error.svelte` documentation in svelte kit

pedrosilva17:
Closes #9
Navbar and footer designs are from [feature/layout](/NIAEFEUP/website-niaefeup-frontend/tree/feature/layout)

error page


toni-santos:
Looking good, but maybe we could make it so that we also receive and display an error message like shown in the docs [here](https://kit.svelte.dev/docs/errors#responses), but we should check up with the image team beforehand maybe. Other than that, it's great! 👍

pedrosilva17:
> Looking good, but maybe we could make it so that we also receive and display an error message like shown in the docs [here](https://kit.svelte.dev/docs/errors#responses), but we should check up with the image team beforehand maybe. Other than that, it's great! +1

Thank you! I thought about displaying error messages with `{$page.error.message}` as well, but I was unsure whether the sentences in the design were just placeholders or an intended generic message for the error page. Something to be discussed in the meeting.

pedrosilva17:
Also, I accidentally added the commits from [feature/layout](/NIAEFEUP/website-niaefeup-frontend/tree/feature/layout) here when pulling the changes from that branch into mine, in order to have the updated page layout. My bad.

BrunoRosendo:
Hey @pedrosilva17, good job, it's looking really nice IMO! Just to warn you that I changed the base branch of the PR to `feature/layout` so we can properly look at the code changes :-)

srissirs (changes requested):
Looks good! But maybe could you center the icon in the background just a little bit? Either that or make both edges touch the navbar and the footer like this? 

image

BrunoRosendo:
> Looks good! But maybe could you center the icon in the background just a little bit? Either that or make both edges touch the navbar and the footer like this?
> 
> image

@srissirs Could you comment on #13 ? The background is being implemented there

src/routes/+error.svelte
```diff
@@ -0,0 +1,23 @@
+<script>
+  import { page } from '$app/stores';
+</script>
+
+<div class="flex flex-col font-source-code-pro justify-center my-20 px-10 gap-20">
```
limwa: Maybe use a `main` tag here

src/routes/+error.svelte
```diff
@@ -0,0 +1,23 @@
...
+        <section class="flex flex-col basis-2/5 text-white text-left break-normal gap-3">
+            <h1 class="text-2xl md:text-5xl font-raleway font-bold">Error type</h1>
+            <h2 class="text-xl md:text-4xl font-medium">Syntax Error</h2>
+        </section>
+        <section class="flex flex-col basis 3/5 text-white text-right break-normal gap-3">
+            <h1 class="text-2xl md:text-5xl font-raleway font-bold">What went wrong?</h1>
+            <h2 class="text-xl md:text-4xl font-medium">There is a semicolon (;) missing somewhere</h2>
+        </section>
+    </div>
+    <button class="flex bg-primary hover:bg-secondary m-auto text-white font-raleway font-bold min-w-50 py-2 px-4 rounded">
```
limwa: Is the functionality of the "Back" button for this PR? If it is, that's missing.

It'd probably be enough to use an `a` tag with `href="/"`

limwa (commented):
Very good work, just two minor suggestions. Let me know what you think! @pedrosilva17

limwa:
Also, @BrunoRosendo, this PR should only be merged when #13 is merged, right?

pedrosilva17:
You're right, that div tag was one of the changes I had in mind after looking at the code again. I wasn't sure if I should add functionality to the button or how it should work (going to the previous page or directly to the home screen), but I can add that too if everyone agrees. Thanks!

limwa (approved):

BrunoRosendo:
> You're right, that div tag was one of the changes I had in mind after looking at the code again. I wasn't sure if I should add functionality to the button or how it should work (going to the previous page or directly to the home screen), but I can add that too if everyone agrees. Thanks!

I can see 2 options regarding the back button:
- The simpler way would be to change it to a "Home" button and just href to "/", I honestly don't think this page will be used that much to bother doing something more complex.
- If we really want a back button, I guess we could look at the browser's history and get the previous URL. If it doesn't exist (first page), maybe go to the home page as well.

src/routes/+error.svelte
```diff
@@ -0,0 +1,23 @@
...
+    </section>
+    <div class="flex flex-row flex-1 m-auto gap-20">
+        <section class="flex flex-col basis-2/5 text-white text-left break-normal gap-3">
+            <h1 class="text-2xl md:text-5xl font-raleway font-bold">Error type</h1>
+            <h2 class="text-xl md:text-4xl font-medium">Syntax Error</h2>
+        </section>
+        <section class="flex flex-col basis 3/5 text-white text-right break-normal gap-3">
+            <h1 class="text-2xl md:text-5xl font-raleway font-bold">What went wrong?</h1>
+            <h2 class="text-xl md:text-4xl font-medium">There is a semicolon (;) missing somewhere</h2>
+        </section>
```
BrunoRosendo: I think these sections could be common enough to generalize (create a component for them), we already repeat a lot of stuff here. The differences seem to be only the text and the text-left/text-right, which could be passed as props

src/routes/+error.svelte
```diff
@@ -0,0 +1,23 @@
...
+    <section class="flex flex-col flex-1 text-white text-center break-normal gap-10">
+        <h1 class="text-7xl md:text-huge font-extrabold">{$page.status}</h1>
+        <h2 class="text-3xl md:text-6xl font-semibold">missing ; before statement</h2>
+    </section>
+    <div class="flex flex-row flex-1 m-auto gap-20">
+        <section class="flex flex-col basis-2/5 text-white text-left break-normal gap-3">
+            <h1 class="text-2xl md:text-5xl font-raleway font-bold">Error type</h1>
+            <h2 class="text-xl md:text-4xl font-medium">Syntax Error</h2>
+        </section>
+        <section class="flex flex-col basis 3/5 text-white text-right break-normal gap-3">
```
BrunoRosendo: The `basis 3/5` might be wrong

src/routes/+error.svelte
```diff
@@ -0,0 +1,23 @@
...
+            <h1 class="text-2xl md:text-5xl font-raleway font-bold">What went wrong?</h1>
+            <h2 class="text-xl md:text-4xl font-medium">There is a semicolon (;) missing somewhere</h2>
+        </section>
+    </div>
+    <a href="/" class="flex bg-primary hover:bg-secondary m-auto text-white font-raleway font-bold min-w-50 py-2 px-4 rounded">
+        <p class="px-5">Voltar</p>
+    </a>
```
BrunoRosendo: I think the button should also be a component, it'll most definitely be used in other places. Adding it to story book is a good idea too (check readme for info)

BrunoRosendo (changes requested):
Good job! Small suggestions

pedrosilva17:
Yikes. I had to rebase this branch with develop to have access to the storybook and used I used `--force-with-lease`, but I accidentally added the whole commit history. My bad. I guess I'll delete this branch and create another one.

pedrosilva17:
After some rebase problems, I applied the suggested changes and the code looks quite better. Some questions:

- Are the locations/names of the components adequate? I though about generalizing the section component even more and putting it in the common/ folder, but I think it can vary too much. Let me know it you think otherwise.
- Initially, the component styles were not being rendered at all, so I added an import on the `preview.js`  file for storybook that enables compatibility with Tailwind. Should I keep it or let it get fixed in another branch?
- Another issue with storybook is that some props are not applied correctly to the component if modified inside its control panel. For example, if I try to apply the color purple to a button on the website or pass it to storybook through the arguments in `Button.stories.js` , it works as intended. However, if the button is of a different color initially and I change it inside storybook, as long as that color does not exist anywhere in the website, it becomes colorless. This also happens with width (even though some values work straight away, like `w-5` and `w-16`), and the rest of the props (href and text) work as intended anywhere they get changed.

What I wanted to accomplish was being able to dynamically modify every button property inside storybook. Although I'm not sure if that's my concern or if I should be implementing variable properties in the first place, I feel like I should let you know. I would attach screenshots of the issues, but currently I do not have storybook in the branch because of the rebase problems I had before. I did save one screenshot of the modifications being applied correctly, though:

image

I'll just wait for `feature/layout` to rebase their branch and then repeat the process properly.

BrunoRosendo:
> After some rebase problems, I applied the suggested changes and the code looks quite better. Some questions:
> 
>     * Are the locations/names of the components adequate? I though about generalizing the section component even more and putting it in the common/ folder, but I think it can vary too much. Let me know it you think otherwise.
> 
>     * Initially, the component styles were not being rendered at all, so I added an import on the `preview.js`  file for storybook that enables compatibility with Tailwind. Should I keep it or let it get fixed in another branch?
> 
>     * Another issue with storybook is that some props are not applied correctly to the component if modified inside its control panel. For example, if I try to apply the color purple to a button on the website or pass it to storybook through the arguments in `Button.stories.js` , it works as intended. However, if the button is of a different color initially and I change it inside storybook, as long as that color does not exist anywhere in the website, it becomes colorless. This also happens with width (even though some values work straight away, like `w-5` and `w-16`), and the rest of the props (href and text) work as intended anywhere they get changed.
> 
> 
> What I wanted to accomplish was being able to dynamically modify every button property inside storybook. Although I'm not sure if that's my concern or if I should be implementing variable properties in the first place, I feel like I should let you know. I would attach screenshots of the issues, but currently I do not have storybook in the branch because of the rebase problems I had before. I did save one screenshot of the modifications being applied correctly, though:
> 
> image
> 
> I'll just wait for `feature/layout` to rebase their branch and then repeat the process properly.

- I think the components should be under a `components/` folder because `lib/` will be used for all kinds of things. I think everything else is good, I also thought about a general section but I'm not sure if it's going to be used
- Here is OK
- Maybe storybook wasn't completely configured with tailwind? Can you try https://storybook.js.org/recipes/tailwindcss

src/lib/common/Button.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts">
+    export let color: string, hover: string, href: string, text: string, width: string;
```
BrunoRosendo: Imo, I don't think we should pass full-fledged tailwind classes as props. For example, typing `width="w-32"` feels really cumbersome, why do have to write width twice?

I'm giving the following suggestions, see if you like them:
- For color it works well because I just type color={color}. Since we use "hover" for a secondary color, I would change the prop to `secondaryColor` or `hoverColor` (or wtv you want) and simply pass the color too. When using the prop, it'd become something like  `hover:{secondaryColor}`
- For width, we could perhaps take on a similar strategy. I also thought about having some semantic prop like "large" or "small" but then I didn't like adding abstraction on top of tailwind
- href shouldn't be mandatory, we don't always want to redirect the user when clicking a button. I'd even consider removing `<a>` from the component but that's your call

src/lib/error/ErrorSection.svelte
```diff
@@ -0,0 +1,8 @@
+<script lang="ts">
+    export let side: string, heading: string, text: string;
```
BrunoRosendo:
```suggestion
    export let align: string, heading: string, text: string;
```

Align seems like a better name and it's consistent with tailwind https://tailwindcss.com/docs/text-align#setting-the-text-alignment

BrunoRosendo (changes requested):
Good job with the refactors! It's looking really good. I left some suggestions regarding prop names and semantics

src/lib/common/Button.svelte
pedrosilva17: I thought about doing this, but they recommend against it in the docs: (https://tailwindcss.com/docs/content-configuration#dynamic-class-names). In fact, I tried to do it now in the width (passing 96 and then using w-{width}) and it didn't work. The solution would be to map keywords to their corresponding class in Tailwind, but since you don't like the abstraction of using "large" and "small", I'm guessing that's not an option.

pedrosilva17: I noticed, however, that I'm doing the exact same thing with the "align" property on the ErrorSection component and it works fine. Strange...

pedrosilva17: Nevermind, I found out why: the first section already uses `text-center` and the footer uses `text-right`, that's why it works when I change any of them. If I change the text-right in the footer to text-left and rerun the start command, the rightmost ErrorSection will be aligned to the left (by default).

BrunoRosendo: I see, I wasn't aware of that (since I'm not experienced with tailwind). I'd suggest following the docs then, since they specifically discourage what I previously suggested

pedrosilva17: Okay. Just to clarify: would you rather have the classes mapped to keywords or write the classes themselves when calling the components? It's a tradeoff between ease of use and versatility, I guess.

BrunoRosendo: You're right, it's a tradeoff. Maybe I'd go for the keywords but it's your call

pedrosilva17:
The layout branch is already up-to-date with develop. Here's what I did to rebase:
- git switch feature/layout
- git pull
- git switch feature/error-page
- git rebase feature/layout
- git push

Didn't work. I tried to revert the rebase but ended up doing more harm than good. Not touching the remote in the meantime. Any way I can do this properly from the start?

BrunoRosendo:
> The layout branch is already up-to-date with develop. Here's what I did to rebase:
> 
>     * git switch feature/layout
> 
>     * git pull
> 
>     * git switch feature/error-page
> 
>     * git rebase feature/layout
> 
>     * git push
> 
> 
> Didn't work. I tried to revert the rebase but ended up doing more harm than good. Not touching the remote in the meantime. Any way I can do this properly from the start?

Your steps are correct but you need to do `git push --force-with-lease` instead of a simple push. You need to force because you changed the commit history when you rebased. "with lease" is used as a safe measure, it does the same as `-f` but aborts if anyone else other than you pushed in the meantime

pedrosilva17:
> > The layout branch is already up-to-date with develop. Here's what I did to rebase:
> > ```
> > * git switch feature/layout
> > 
> > * git pull
> > 
> > * git switch feature/error-page
> > 
> > * git rebase feature/layout
> > 
> > * git push
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Didn't work. I tried to revert the rebase but ended up doing more harm than good. Not touching the remote in the meantime. Any way I can do this properly from the start?
> 
> Your steps are correct but you need to do `git push --force-with-lease` instead of a simple push. You need to force because you changed the commit history when you rebased. "with lease" is used as a safe measure, it does the same as `-f` but aborts if anyone else other than you pushed in the meantime

You're right, I forgot about --force-with-lease. I think I was able to recover the original branch and rebased properly again, the log looks like this:

image

And here's the git reflog:

image


I'm just double-checking with you to avoid another unnecessary accident. If everything is in order, I'll push the rebase and then work on the proper storybook integration.

pedrosilva17:
Looking at it again, I noticed a December 22nd commit on top of the more recent ones and a suspicious merge. I'll wait for the issue to be resolved on the other branch before making changes. (Edit: I think the issue here is the merge actually, the other 2 commits make sense because they don't appear at the bottom of the log)

BrunoRosendo:
> Looking at it again, I noticed a December 22nd commit on top of the more recent ones and a suspicious merge. I'll wait for the issue to be resolved on the other branch before making changes. (Edit: I think the issue here is the merge actually, the other 2 commits make sense because they don't appear at the bottom of the log)

You might want to check between you 2/3

pedrosilva17:
> > Looking at it again, I noticed a December 22nd commit on top of the more recent ones and a suspicious merge. I'll wait for the issue to be resolved on the other branch before making changes. (Edit: I think the issue here is the merge actually, the other 2 commits make sense because they don't appear at the bottom of the log)
> 
> You might want to check between you 2/3

I already talked with António, and decided to leave this issue on hold until they merge their issue with develop. I will then rebase with develop and merge my changes there. Otherwise, there's a lot of back-and-forth done which will only cause more headaches. There's not a lot left to do anyways (just properly setting up storybook), so I don't think there's a problem.

srissirs (approved):

thePeras:
Hey @pedrosilva17, the text looks huge in my screen

Screenshot 2023-02-19 at 19 32 11

pedrosilva17:
> Hey @pedrosilva17, the text looks huge in my screen
> 
> Screenshot 2023-02-19 at 19 32 11

It does, can you leave your resolution here so I can test it later? I'll fix it asap

thePeras:
> 
> It does, can you leave your resolution here so I can test it later? I'll fix it asap

1440x900

pedrosilva17:
> > It does, can you leave your resolution here so I can test it later? I'll fix it asap
> 
> 1440x900

I changed the size breakpoints a bit and now it looks like this on your resolution:

image

What do you think?

thePeras:
> What do you think?

Looks good

BrunoRosendo:
@pedrosilva17 Can you update the branch?

pedrosilva17:
Branch updated. There are still some linting issues, but I'll take care of those in a bit

BrunoRosendo (approved):

PR #34, opened 2023-01-11
Fix CI audit

limwa:
This pull request intends to fix the project so that the audit CI passes.

# Review checklist

- [ ] Behavior is as expected


limwa:
Closing this pull request since the vulnerabilities are present on Storybook dependencies. We still have to decide if we are going to just ignore the CI fails, remove Storybook altogether or do something else.

PR #38, opened 2023-01-13
feat: initial work on hexagons

### LINKED ISSUE #1: Hexagons
Hexagon component that will be used on many pages across the website. Some examples follow below:

image
image

limwa:
Closes #1 

This PR adds two components, called `HexagonGrid` and `Hexagon`. The `HexagonGrid` is responsible for the placement of the elements and the `Hexagon` is responsible for formatting it's content according to an hexagon shape.

The `HexagonGrid` receives a number of columns and a list of items and, then, creates the required elements in the necessary positions. Each element can access a row, column and the corresponding item.

# Screenshots

HexagonGrid with dark gray Hexagons inside

# Review checklist

- [x] Behavior is as expected
- [x] Clean, well structured code


limwa:
A problem I'm having currently is that the background in between the white outlines is solid instead of transparent. This is not so easily fixable, I will have to export a SVG to use and overlap it with the current hexagons and make sure that both hexagons are of the same size.

Currently, the HexagonImage doesn't adapt very well to different screen resolutions (will be fixed).

limwa:
> Currently, the HexagonImage doesn't adapt very well to different screen resolutions (will be fixed).

For instance, at 200% zoom, it looks like this:
image

limwa:
I have updated the pull request description with a new image of the `HexagonGrid` with `Hexagons`. The functionality where the members' photos is displayed inside will be taken care of in another PR.

limwa:
No preview for this PR is available and we are experiencing some problems with Storybook, so the only way to get hands-on with the hexagons is to checkout the branch and put the following code on `+page.svelte`.

```svelte
<script>
    import Hexagon from "@/components/Hexagons/Hexagon.svelte";
    import HexagonGrid from "@/components/Hexagons/HexagonGrid.svelte";
</script>

<div class="w-3/4 mx-auto mt-20">
    <HexagonGrid items={Array(100).fill(0)} cols={5} gap="medium">
        <Hexagon>
            <div class="h-full w-full bg-slate-900"></div>
        </Hexagon>
    </HexagonGrid>
</div>
```

src/components/Hexagons/Hexagon.svelte
```diff
@@ -0,0 +1,29 @@
...
+    mask-image: url('/masks/hexagon/background.svg');
+  }
+
+  .clip-outer-hexagon {
+    padding: 5.8% 6.9%; /* Magic Numbers */
+    mask-image: url('/masks/hexagon/foreground.svg');
+  }
+
+  .clip-hexagon {
+    clip-path: polygon(0% 50%, 25% 100%, 75% 100%, 100% 50%, 75% 0%, 25% 0%);
```
BrunoRosendo: Did you consider using css var to avoid repeating the polygon values?

src/components/Hexagons/HexagonGrid.svelte
```diff
@@ -0,0 +1,44 @@
...
+  // Inspired by https://github.com/sveltejs/svelte-virtual-list/blob/master/VirtualList.svelte
+  type T = $$Generic;
+
+  export let items: T[];
+  export let cols: number;
+
+  export let gap: 'small' | 'medium' | 'big' = 'medium';
+</script>
+
+<!-- Warning: Very magic, totally eyeballed numbers in the next line -->
```
BrunoRosendo: warning makes sense, maybe a little less goofy xd

src/components/Hexagons/HexagonGrid.svelte
```diff
@@ -0,0 +1,44 @@
...
+>
+  {#each items as item, index}
+    {@const col = index % cols}
+    {@const row = (index - col) / cols}
+    {@const colStart = 2 * col + 1}
+    {@const colEnd = 2 * col + 4}
+    {@const rowStart = 2 * row + 1 + (col % 2)}
+    {@const rowEnd = 2 * row + 3 + (col % 2)}
+    <li style="grid-column: {colStart} / {colEnd}; grid-row: {rowStart} / {rowEnd}">
+      <slot {item} {col} {row} />
```
BrunoRosendo: I'm trying to thing how the slot can use the col and row, do you have an example?

BrunoRosendo (commented):
I don't have that much to say, I thought it was going to be way harder to review but it turned our rather simple so good job on that. I just left very minor comments

src/components/Hexagons/Hexagon.svelte
limwa: I changed the classes in the Hexagon.svelte file to be more tailwind-like!

src/components/Hexagons/HexagonGrid.svelte
limwa: Okay, I've changed the comment. Let me know what you think!

src/components/Hexagons/HexagonGrid.svelte
limwa: I was thinking maybe conditional styling based on the column or row. Or even skipping the rendering of certain items based on the column and/or row. Do you think I should remove that?

toni-santos: This might be useful, i believe it will mostly depend on how we use the hexagons however. We could keep it for now and if we find it to not be useful we could then remove it.

toni-santos:
This is looking great! :)
We currently have two `components` folders, one in ` src/lib/component` and another one in `src/lib/components`.
Other than that, I think we should talk with the design team in order evaluate if this implementation is adequate, keeping in mind the fact that there are some different hexagons implementations designed.

limwa:
About that, today I we noticed during the meeting that the vertical hexagons weren't implemented yet. Gonna have to take some time on that.

limwa:
So, I have a few questions now that we have horizontal and vertical hexagons

1. Do we keep the Hexagon component? Currently, it only acts as a switch between displaying the VerticalHexagon and the HorizontalHexagon components
2. Should I wrap the slot in the HexagonGrid with a the corresponding Hexagon component? The grid spacings are not very compatible with other components, so I'd like to hear your opinions on this.
3. Do you have any tips on how to make the code more organized? xD

@BrunoRosendo @toni-santos

src/components/Hexagons/HexagonGrid.svelte
BrunoRosendo: I think that makes sense but it's not super clear, maybe a comment would help

BrunoRosendo:
> So, I have a few questions now that we have horizontal and vertical hexagons
> 
> 1. Do we keep the Hexagon component? Currently, it only acts as a switch between displaying the VerticalHexagon and the HorizontalHexagon components
> 2. Should I wrap the slot in the HexagonGrid with a the corresponding Hexagon component? The grid spacings are not very compatible with other components, so I'd like to hear your opinions on this.
> 3. Do you have any tips on how to make the code more organized? xD
> 
> @BrunoRosendo @toni-santos

1.  I think the abstraction makes sense, nobody needs to know the internal implementation
2. Doesn't that go against the implementation of components like `ProfileHexagon`? If that's not a huge issue then I'd prefer injecting the items directly on the grid instead of having to call `HexagonGrid` and `Hexagon`, seems redundant since I'm already expecting an hexagon grid
3. I'll try to leave suggestions when/if it's already done

BrunoRosendo:
> This is looking great! :) We currently have two `components` folders, one in ` src/lib/component` and another one in `src/lib/components`. Other than that, I think we should talk with the design team in order evaluate if this implementation is adequate, keeping in mind the fact that there are some different hexagons implementations designed.

We should rethink the way we're organizing files under `lib/`, we can either organize them based on features (e.g. `lib/auth`, `lib/events`) while including components and business logic, or based on their utility (e.g. `lib/components/hexagons/`, `lib/components/mainPage`, `/lib/services/...`, `/lib/interfaces/...`). I like the second approach more but we don't need to make the change in this PR

limwa:
> > So, I have a few questions now that we have horizontal and vertical hexagons
> > 
> > 1. Do we keep the Hexagon component? Currently, it only acts as a switch between displaying the VerticalHexagon and the HorizontalHexagon components
> > 2. Should I wrap the slot in the HexagonGrid with a the corresponding Hexagon component? The grid spacings are not very compatible with other components, so I'd like to hear your opinions on this.
> > 3. Do you have any tips on how to make the code more organized? xD
> > 
> > @BrunoRosendo @toni-santos
> 
> 1. I think the abstraction makes sense, nobody needs to know the internal implementation
> 2. Doesn't that go against the implementation of components like `ProfileHexagon`? If that's not a huge issue then I'd prefer injecting the items directly on the grid instead of having to call `HexagonGrid` and `Hexagon`, seems redundant since I'm already expecting an hexagon grid
> 3. I'll try to leave suggestions when/if it's already done

2. Good point, I hadn't thought about using other Hexagons inside of the HexagonGrid. I know what you mean when you say it "seems redundant", but, this way, we can have different types of hexagons, like EventHexagon, ProfileHexagon that will be injected into the grid. Another way of doing it would be something like `<HexagonGrid component={ProfileHexagon} items={...} cols={...} />`, however, this would mean that the hexagons wouldn't be able to receive props, so I'm not a big fan of this solution. I think I'll keep the need to use the Hexagon inside of the HexagonGrid.

3. The code is finished, but it has a lot of repetition. I don't know how to organize it better because many things from the horizontal and vertical versions are very similar but not quite the same :/

limwa:
> > This is looking great! :) We currently have two `components` folders, one in ` src/lib/component` and another one in `src/lib/components`. Other than that, I think we should talk with the design team in order evaluate if this implementation is adequate, keeping in mind the fact that there are some different hexagons implementations designed.
> 
> We should rethink the way we're organizing files under `lib/`, we can either organize them based on features (e.g. `lib/auth`, `lib/events`) while including components and business logic, or based on their utility (e.g. `lib/components/hexagons/`, `lib/components/mainPage`, `/lib/services/...`, `/lib/interfaces/...`). I like the second approach more but we don't need to make the change in this PR

I like the second approach the most as well. Inside those utility folders, we could create per-feature folders.
I agree that we don't need to make the change in this PR.

toni-santos:
I really like how this is looking! I don't have much to point out except from the fact that it feels a bit weird to need to pass the orientation to an Hexagon when we already give it to the hexagon grid, it would be nice if we were just able to pass it to one or the other, although I understand why it is being done this way.
Regardless I am happy with this PR in its current state and we could maybe merge it and in the future, once some pages that use these components have some more advanced work being done, tweak any issues with this.

toni-santos (approved):

limwa:
I would like to comment on the need to use the orientation twice.

The orientation is needed on the grid because the grid layout changes according to the orientation. Vertical hexagons need to have a bit more vertical spacing and horizontal hexagons need to have a bit more horizontal space.

The orientation is also needed on the hexagon itself because it needs to know whether to rotate or not.

To solve this problem, we'd need to either:

1. Set the orientation on the grid and the grid transmits the information to the hexagons
2. Set the orientation on the hexagons and the hexagons transmit the information to the grid

Solution 1 is not viable, because some hexagons (such as the profile hexagon) can only be displayed horizontal and, as such, the orientation can't be controlled by the grid. Furthermore, it should be possible to display hexagons outside of a grid, therefore, they'd need to know their orientation even in those situations.

Solution 2 is the only one but passing data from a child to a parent is very hard. We could also try setting a const on the component and see if it can be read in some way. I have no idea how to do it in svelte. I'll search for some ideas and let you know if I find something out.

src/lib/layout/hexagons/HexagonGrid.svelte
```diff
@@ -0,0 +1,74 @@
...
+      {@const col = index % cols}
+      {@const row = (index - col) / cols}
+      {@const rowStart = 2 * row + 1}
+      {@const rowEnd = 2 * row + 4}
+      {@const colStart = 2 * col + 1 + (row % 2)}
+      {@const colEnd = 2 * col + 3 + (row % 2)}
+      <li style="grid-column: {colStart} / {colEnd}; grid-row: {rowStart} / {rowEnd}">
+        <slot {item} {col} {row} />
+      </li>
+    {/if}
```
DoStini: I'm not sure if you will agree with me, but I don't enjoy the duplicated code.
I propose something like this, but I can agree to an extent that the code is a little less readable, but most of this are magic numbers so I think this might be better.
```suggestion
{#each items as item, index}

    {@const isHorizontal = orientation === 'horizontal' }
    {@const col = index % cols}
    {@const row = (index - col) / cols}
    {@const horizontalColStart = 2 * col + 1}
    {@const colStart = horizontalColStart + (isHorizontal ? 0 : (row % 2))}
    {@const verticalRowStart = 2 * row + 1}
    {@const rowStart = verticalRowStart + (isHorizontal ? (col % 2) : 0)}
    {@const colEnd = isHorizontal ? 2 * col + 4 : 2 * col + 3 + (row % 2)}
    {@const rowEnd = isHorizontal ? 2 * row + 3 + (col % 2) : 2 * row + 4}

    <li style="grid-column: {colStart} / {colEnd}; grid-row: {rowStart} / {rowEnd}">
        <slot {item} {col} {row} />
    </li>
```

I think that at least col and row can be extracted from inside elses because they are the same in both conditions, but the rest I leave for your description

svelte.config.js
```diff
@@ -14,7 +14,10 @@ const config = {
   ],
 
   kit: {
-    adapter: adapter()
+    adapter: adapter(),
+    alias: {
+      '@/*': './src'
+    }
```
DoStini, line 20: Just curiosity, what is this for?

DoStini (approved):

src/lib/layout/hexagons/HexagonGrid.svelte
limwa: Thanks for the help! I'll keep looking for other ways to remove code repetition in other files as well.

svelte.config.js
limwa: It was an import alias that I used during development. It's use is shown [here](https://github.com/NIAEFEUP/website-niaefeup-frontend/pull/38#issuecomment-1483869247https://github.com/NIAEFEUP/website-niaefeup-frontend/pull/38#issuecomment-1483869247).
However, now that the components folder is inside `lib/`, we can simply use `$lib` instead of the `@`.

src/lib/layout/hexagons/HexagonGrid.svelte
```diff
@@ -0,0 +1,68 @@
...
+  const gridColumnsStyle =
+    orientation === 'horizontal'
+      ? `grid-template-columns: 0.2425fr repeat(${cols}, 0.4805fr 0.2425fr); grid-auto-rows: 0.5fr 0.5fr;`
+      : `grid-template-columns: 0.5fr repeat(${cols}, 0.505fr 0.5fr); grid-auto-rows: 0.251fr 0.5fr;`;
+</script>
+
+<!-- The grid column sizes used are magic numbers that look good with the Hexagon component -->
+<ul class="grid gap-{orientation}-{gap}" style={gridColumnsStyle}>
+  {#each items as item, index}
+{#each items as item, index}
```
DoStini:
```suggestion
  {#each items as item, index}
```

DoStini (changes requested):

DoStini:
I have left a suggestion under feature/hexagons-example to pass the component as a prop instead of using slot, so that we can better control which components can be used with the grid. Let me know what you think

BrunoRosendo:
@limwa lint please

DoStini:
Currently we can create an hexagon grid with the following:

```svelte
<HexagonGrid 
    orientation="vertical" 
    gap="big" 
    items={[1,2,3,4,5,6,7,8,9]}
    cols={5}
    component={RedExampleHexagon} 
>
</HexagonGrid>
<HexagonGrid 
    orientation="vertical" 
    gap="big" 
    items={[1,2,3,4,5,6,7,8,9]}
    cols={5}
    component={BlueExampleHexagon} 
>
</HexagonGrid>
 ```

Left the component examples and should be removed after we create an actual component

**Comment by BrunoRosendo** [comment]
> Currently we can create an hexagon grid with the following:
> 
> ```svelte
> <HexagonGrid 
>     orientation="vertical" 
>     gap="big" 
>     items={[1,2,3,4,5,6,7,8,9]}
>     cols={5}
>     component={RedExampleHexagon} 
> >
> </HexagonGrid>
> <HexagonGrid 
>     orientation="vertical" 
>     gap="big" 
>     items={[1,2,3,4,5,6,7,8,9]}
>     cols={5}
>     component={BlueExampleHexagon} 
> >
> </HexagonGrid>
> ```
> 
> Left the component examples and should be removed after we create an actual component

That can probably be done in #127

File: src/lib/layout/hexagons/HexagonGrid.svelte [hunk 7]
```diff
@@ -17,21 +21,19 @@
...
+    {@const rowStart = verticalRowStart + (isHorizontal ? col % 2 : 0)}
+    {@const colEnd = isHorizontal ? 2 * col + 4 : 2 * col + 3 + (row % 2)}
+    {@const rowEnd = isHorizontal ? 2 * row + 3 + (col % 2) : 2 * row + 4}
 
-      <li style="grid-column: {colStart} / {colEnd}; grid-row: {rowStart} / {rowEnd}">
-        <slot {item} {col} {row} />
-      </li>
-    {/each}
+    <li style="grid-column: {colStart} / {colEnd}; grid-row: {rowStart} / {rowEnd}">
+      <svelte:component this={component} {orientation} data={item} />
```
BrunoRosendo, line 35: pretty nice!

BrunoRosendo (approved):

bdmendes (approved):

BrunoRosendo:
@DoStini we need your approval to merge this

limwa:
@BrunoRosendo now, we don't 😜

DoStini (approved):

PR #47, opened 2023-01-27
Feature/docker

### LINKED ISSUE #3: global: dockerize the application
To ease future deployments, we should be able to run our application using containers.

We should create a Dockerfile for development, testing and production

jamcunha:
Closes #3 

Dockerized the frontend.

To use the dev environment do
```
docker-compose build website-niaefeup-frontend-dev
docker-compose up website-niaefeup-frontend-dev
```
Run tests with `docker-compose run --rm website-niaefeup-frontend-dev npm run test`.

To use the storybook do 
```
docker-compose build website-niaefeup-frontend-storybook
docker compose up website-niaefeup-frontend-storybook
```

# Review checklist

- [ ] Behavior is as expected
- [ ] Clean, well structured code


BrunoRosendo (changes requested):
Good job so far! It'd be useful to have a dockerfile to run the tests. 
Also, the lint is failing. You can fix it automatically, the info is in the readme

jamcunha:
Should I remove production in docker-compose? I only left that to be easier to test.

BrunoRosendo:
> Should I remove production in docker-compose? I only left that to be easier to test.

I think it's ok

Dockerfile.dev
```diff
@@ -0,0 +1,24 @@
...
+RUN mkdir -p /usr/src/website-niaefeup-frontend
+WORKDIR /usr/src/website-niaefeup-frontend
+
+COPY package.json ./
+COPY package-lock.json ./
+
+RUN npm install
+
+COPY .storybook/ .storybook/
+COPY .svelte-kit/ .svelte-kit/
```
BrunoRosendo: This folder is not part of the repo so it shouldn't be copied. If I build the image without it, I get an error. I think you can solve this by removing the COPY and running npm install after copying everything else

BrunoRosendo: Same for other dockerfiles

Dockerfile.prod
```diff
@@ -0,0 +1,36 @@
+FROM node:19.5.0-alpine AS build
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+COPY package.json ./
+COPY package-lock.json ./
+
+ENV NPM_CONFIG_COLOR=false
```
BrunoRosendo, line 6: I'm curious, what does this env do?

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      dockerfile: Dockerfile.storybook
+    ports:
+      - 6006:6006
+    volumes:
+      - ./src:/usr/src/website-niaefeup-frontend/src
+      - ./static:/usr/src/website-niaefeup-frontend/static
+    environment:
+      - PORT=6006
+
+  website-niaefeup-frontend-test:
```
BrunoRosendo, line 27: test image works well but I have to press ctrl+c after the tests execute. The test run is expecting me to press "q" to quit but the docker shell is not interactive. I think we could either start an interactive shell (`-it` flag with docker so there should be a way to do it with docker-compose) or check if there's a flag in vitest that exits automatically (less preferred because vitest detects changes and we might want the feature)

package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "start": "npm run dev",
-    "dev": "vite dev --port 3000",
+    "start": "node ./build",
```
BrunoRosendo: I really enjoy just running `npm start` instead of `npm run dev`, could you leave this script and add a new one (e.g. prod) for this?

BrunoRosendo (changes requested):
I tried running the containers and my machines and had an issue here and there. They should be easy fixed, so it's almost done!

On another note, I'd personally rename Dockerfile.dev to Dockerfile to ease commands to someone using docker (instead of docker-compose).

bdmendes (commented):
Can we copy [all directory contents](https://stackoverflow.com/questions/65720449/in-dockerfile-copy-all-contents-of-current-directory-except-one-directory) in a one-liner and explicitly ignore all unintended copies? That would make this more future-proof in case we add more config files later.

Dockerfile.prod
jamcunha: I saw that in nijobs Dockerfile and decided to put it in. It's supposed to disable colours in order not to break logs.

BrunoRosendo (approved):
looks good!

bdmendes (approved):

Dockerfile.prod
```diff
@@ -20,6 +20,5 @@ COPY --from=build /usr/src/app/package-lock.json ./package-lock.json
 COPY --from=build /usr/src/app/build build/
 RUN npm install --omit=dev
 
-ENV PORT=80
```
BrunoRosendo: The idea is to leave the PORT env var to be defined in the server?

BrunoRosendo (changes requested):
I had already approved but, after seeing the README changes, I think we could explain how to run w/ docker and npm. Some people could prefer running with npm

Dockerfile.prod
jamcunha: That is the idea, but if it doesn't matter which port is chosen, we can have it as it was before.

BrunoRosendo (approved):

PR #108, opened 2023-03-11
Notification snackbars

### LINKED ISSUE #16: global: create snackbar
We should have a reusable snackbar that can be used on any part of the website as shown in the mockups and screenshot below.

These should have arbitrary text, stay for a defined amount of time (or optionally passed as a parameter) before disappearing, and have an "X" button enabling the user to close it earlier.

Please make sure to handle situations where the snackbars can stack (multiple notifications at the same time). A possible option is having a maximum number of notifications stack on each other and after that is reached, either wait for a notification to close before displaying another or close the oldest one automatically.

Consider using an external library for this functionality.

image

BrunoRosendo:
closes #16 

This PR adds support for displaying notifications. There's currently a temporary button to generate notifications so you can test it. I'll remove it if you think they're ok.

# Screenshots

image

# Review checklist

- [ ] Behavior is as expected


src/lib/notifications/Snackbar.svelte
```diff
@@ -0,0 +1,28 @@
...
+  import type { Notification } from '.';
+
+  export let notification: Notification;
+</script>
+
+<div
+  out:slide
+  in:fly={{ x: 100 }}
+  class="mb-4 flex h-auto w-80 rounded-lg bg-primary px-4 py-2"
+  role="alert"
```
pedrosilva17: Maybe we shouldn't use `role="alert"` for every single snackbar notification, but rather `role="status"` for the majority of them.

Here's what I've read on the subject:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role
"The alert role is used to communicate an important and usually time-sensitive message to the user."
"Warning: Because of its intrusive nature, the alert role must be used sparingly and only in situations where the user's immediate attention is required."

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role
"A status is a type of [live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) providing advisory information that is not important enough to justify an alert, which would immediately interrupt the announcement of a user's current activity."

If we aren't using notifications to only display very important information, then the status role is a more adequate default value, as it will not cause screen readers to interrupt their current activity, which is something that can disorient some users.
In my opinion, leaving it with `role="status"` should work for now. In the future, adding a more urgent variation of a notification which uses the alert role and no timeout (or a longer one) could be an option. 

What do you think?

pedrosilva17 (commented):
Looks very solid! I left a comment on the use of `role="alert"`, but the rest looks clean. Let me know if you agree with it 👍

src/lib/notifications/Snackbar.svelte
BrunoRosendo: I definitely agree with you! I wasn't aware of that difference, I'll make the change. Thanks!

pedrosilva17 (approved):
LGTM 🚀

marineve17 (approved):
Very clean design and animations, good job!

thePeras (changes requested):
Don't center the close icon, fix it at the top

marineve17:
I think this would look even better with a progress bar on the bottom, showing the time until the notification auto-closes. However, this depends on the kind of notifications we will be sending on the website. If it's something the user should read, there should be a progress bar, if not, something like "E-mail copied", I don't think there's any need to change this. What do you think @srissirs @thePeras?

srissirs:
> I think this would look even better with a progress bar on the bottom, showing the time until the notification auto-closes. However, this depends on the kind of notifications we will be sending on the website. If it's something the user should read, there should be a progress bar, if not, something like "E-mail copied", I don't think there's any need to change this. What do you think @srissirs @thePeras?

If it’s an action that takes some time to load yes, I believe a load bar is pertinent, but, if not, I don’t think it’s necessary. I’m not sure if you meant the load bar should appear below/ nearby the notification pop up, but if we consider for example the loading of an image , I believe it would be better if the loading bar appeared where you clicked the button to upload the image. I think it might be confusing for the user otherwise. 
Please let me know if I misunderstood your suggestion.

thePeras:
I don't think a progress bar at the bottom of the notification is really necessary. It is intuitive that a snack bar will disappear. 
The only disadvantage I see with that is it creates more tension on the user focus to the right bottom of the screen.
I prefer a more sublime approach.

marineve17:
> If it’s an action that takes some time to load yes, I believe a load bar is pertinent, but, if not, I don’t think it’s necessary. I’m not sure if you meant the load bar should appear below/ nearby the notification pop up, but if we consider for example the loading of an image , I believe it would be better if the loading bar appeared where you clicked the button to upload the image. I think it might be confusing for the user otherwise. Please let me know if I misunderstood your suggestion.

I think you misunderstood? I meant something similar to [react-toast](https://fkhadra.github.io/react-toastify/introduction/). If you click on "Show toast", there's a rainbow progress bar under the notification.

BrunoRosendo:
I changed the close button's position but I didn't implement any progress bar since no conclusion has been made

thePeras (changes requested):
The position is perfect. Can you just add a little of margin between the icon and the text so that when the button is hover It's not so close?

Screenshot 2023-03-20 at 09 36 24

BrunoRosendo:
> The position is perfect. Can you just add a little of margin between the icon and the text so that when the button is hover It's not so close?
> 
> Screenshot 2023-03-20 at 09 36 24

It looks pretty different in my screen so please check if the margin I added is enough

thePeras (approved):
✅ 🛫

PR #127, opened 2023-03-25
TeamMemberHexagon component

### LINKED ISSUE #89: team: Hexagon for members
After #38 , we need to start implementing hexagons for different pages. The first one could be the members page, as shown below.
This work will later be used in #19. Please also include the new component in storybook.

image

The darker background is the on:hover event

MRita443:
Closes #89 

Developed HexagonProfile component, meant to be the content of each hexagon in the member page

# Screenshots

# Review checklist

- [ ] Behavior is as expected


MRita443:
I am indecisive on which icon to use when a customWebsite is not provided with an icon asset, here are three options I tried:

image

image

image

Do you guys have a preference?

BrunoRosendo:
> I am indecisive on which icon to use when a customWebsite is not provided with an icon asset, here are three options I tried:
> 
> image
> 
> image
> 
> image
> 
> Do you guys have a preference?

imo the second one fits best

LuisDuarte1 (commented):
Please update your branch first before reviewing since #164 has been merged and I think that it might have some conflicts (and maybe you have to make some changes) relative to yours ^^

MRita443:
Since this component is made specifically was made specifically with the team page in mind and its grid is horizontal, do you guys think it's important this component also works for vertical hexagons?

LuisDuarte1:
> Since this component is made specifically was made specifically with the team page in mind and its grid is horizontal, do you guys think it's important this component also works for vertical hexagons?

The hexagon grid on mobile on the team page is horizontal? And also see if there are situations that the TeamMemberHexagon could be used on other pages (eg.: members on a project) and on desktop/mobile it is not vertical

MRita443:
> > Since this component is made specifically was made specifically with the team page in mind and its grid is horizontal, do you guys think it's important this component also works for vertical hexagons?
> 
> The hexagon grid on mobile on the team page is horizontal? And also see if there are situations that the TeamMemberHexagon could be used on other pages (eg.: members on a project) and on desktop/mobile it is not vertical

I just checked, in all the situations the hexagons are always horizontal

src/lib/layout/hexagons/TeamMemberHexagon.svelte
```diff
@@ -0,0 +1,64 @@
...
+  import Hexagon from './Hexagon.svelte';
+  import Icon from '$lib/component/Icon.svelte';
+  import Icons from '$lib/component/Icons';
+  import type { TeamMember } from '@/model/TeamMember';
+
+  export let orientation: 'horizontal' | 'vertical';
+  export let teamMember: TeamMember;
+</script>
+
+<Hexagon {orientation}>
```
LuisDuarte1: If vertical hexagons don't work properly and are not needed maybe remove them from being an option to avoid future mistakes :smile: maybe you can create a low priority issue (not sure if it is beginner friendly) to address vertical orientations when those are needed ^^

LuisDuarte1 (changes requested):
Good job :rocket: just have a suggestion below and maybe fix the storybook accessibility issues ^^

MRita443:
On mobile, the socials icons appear very small. I have tried several different size units but I still haven´t found one that looks proper on mobile. Suggestions are appreciated ^^

Desktop:
image

Mobile:
image

src/lib/layout/hexagons/TeamMemberHexagon.svelte
MRita443: Will work on the accessibility issues next!

src/lib/layout/hexagons/TeamMemberHexagon.svelte
```diff
@@ -0,0 +1,77 @@
...
+            <Icon src={Icons.Linkedin} color="white" size="2.5vw" /></a
+          >
+        {/if}
+        {#if teamMember.gitHub}
+          <a
+            href={teamMember.gitHub}
+            class="opacity-0 transition-all duration-500 ease-out group-hover:opacity-100"
+            aria-label="{teamMember.name}'s GitHub"
+            ><Icon src={Icons.Github} color="white" size="2.5vw" /></a
+          >
```
pedrosilva17: Using viewport width here does make the icons _really_ small on mobile. We could try to use percentage-based widths, but since the icon's parent is the anchor tag, it's not going to make a difference. 
However, what if you don't try to modify the size of the icon, but instead change its parent's size and pass "100%" to the icon's width? Like so:
```suggestion
        {#if teamMember.gitHub}
          <a
            href={teamMember.gitHub}
            class="w-[20%] opacity-0 transition-all duration-500 ease-out group-hover:opacity-100"
            aria-label="{teamMember.name}'s GitHub"
            ><Icon src={Icons.Github} color="white" size="100%" /></a
          >
```

What this does is adapt the width of the anchor tag to 20% of the width of the parent div around it (the horizontal container holding all of the icons) and increase the width of the icon to occupy the whole anchor tag. This will preserve the ratio between container/icon sizes and make the icons a lot bigger at smaller viewports. Here's how it looks when applied to all icons:

Desktop
image

Small Mobile
image

Very Small Viewport
image

It seems to work fine with the specified percentages, but feel free to experiment a bit more.

src/lib/layout/hexagons/TeamMemberHexagon.svelte
```diff
@@ -0,0 +1,77 @@
...
+        text-center text-xs leading-tight text-gray-100 opacity-0 transition-all duration-500 ease-out group-hover:opacity-100 sm:text-xs md:text-sm lg:text-base xl:text-lg"
+      >
+        {teamMember.role}
+      </p>
+      <div class="mt-1 flex justify-center space-x-1">
+        {#if teamMember.linkedin}
+          <a
+            href={teamMember.linkedin}
+            class="opacity-0 transition-all duration-500 ease-out group-hover:opacity-100"
+            aria-label="{teamMember.name}'s LinkdIn"
```
pedrosilva17: Typo
```suggestion
            aria-label="{teamMember.name}'s LinkedIn"
```

src/lib/layout/hexagons/TeamMemberHexagon.svelte
```diff
@@ -0,0 +1,77 @@
...
+        {#if teamMember.customWebsites}
+          {#each teamMember.customWebsites as customWebsite}
+            <a
+              href={customWebsite.url}
+              class="opacity-0 transition-all duration-500 ease-out group-hover:opacity-100"
+            >
+              {#if customWebsite.iconPath}
+                <img
+                  src={customWebsite.iconPath}
+                  alt="Icon of of a custom social website"
```
pedrosilva17: Typo
```suggestion
                  alt="Icon of a custom social website"
```

pedrosilva17 (changes requested):
Great job! I left some small corrections and a suggestion regarding the icon sizes.

src/lib/layout/hexagons/TeamMemberHexagon.svelte
```diff
@@ -0,0 +1,77 @@
...
+            aria-label="{teamMember.name}'s GitHub"
+            ><Icon src={Icons.Github} color="white" size="2.5vw" /></a
+          >
+        {/if}
+        {#if teamMember.customWebsites}
+          {#each teamMember.customWebsites as customWebsite}
+            <a
+              href={customWebsite.url}
+              class="opacity-0 transition-all duration-500 ease-out group-hover:opacity-100"
+            >
```
pedrosilva17: If you remove the iconPath from the storybook example, accessibility issues are raised.
image

It needs an aria-label attribute (you probably just forgot, since you added it on the other icons):
```suggestion
            <a
              href={customWebsite.url}
              class="opacity-0 transition-all duration-500 ease-out group-hover:opacity-100"
              aria-label="{teamMember.name}'s custom website"
            >
```

pedrosilva17 (dismissed):
Forgot to add this, sorry!

src/lib/layout/hexagons/TeamMemberHexagon.svelte
MRita443: This worked really well, thank you for the suggestion!

MRita443:
I have two questions I'd like to discuss with you guys:

Based on these mockups, I tried to place first name and last name on different lines.

image

      <p
        class="mx-auto text-center text-sm font-bold leading-tight text-gray-100 transition-all sm:text-sm md:text-sm lg:text-base xl:text-xl">
        {teamMember.name.split(' ')[0]}
        <br />
        {teamMember.name.split(' ')[1]}
      </p>`

However, what should happen when someone enters more than two names?

image

Do you guys think we should stick to the mockups in this matter and if so, do you have any suggestions for longer names?


The second question is related to the hover state on mobile, as that is obviously not possible. I was thinking of replacing it by click, but I've looked into a couple of things and I don't know how to have that effect for mobile only. Suggestions would be appreciated.

MRita443:
> I have two questions I'd like to discuss with you guys:
> 
> Based on these mockups, I tried to place first name and last name on different lines.
> 
> image
> 
> ```
>   <p
>     class="mx-auto text-center text-sm font-bold leading-tight text-gray-100 transition-all sm:text-sm md:text-sm lg:text-base xl:text-xl">
>     {teamMember.name.split(' ')[0]}
>     <br />
>     {teamMember.name.split(' ')[1]}
>   </p>`
> ```
> 
> However, what should happen when someone enters more than two names?
> 
> image
> 
> Do you guys think we should stick to the mockups in this matter and if so, do you have any suggestions for longer names?
> 
> The second question is related to the hover state on mobile, as that is obviously not possible. I was thinking of replacing it by click, but I've looked into a couple of things and I don't know how to have that effect for mobile only. Suggestions would be appreciated.

Came up with a working version for the second problem, but more concise suggestions would be appreciated

BrunoRosendo:
> I have two questions I'd like to discuss with you guys:
> 
> Based on these mockups, I tried to place first name and last name on different lines.
> 
> image
> 
> ```
>   <p
>     class="mx-auto text-center text-sm font-bold leading-tight text-gray-100 transition-all sm:text-sm md:text-sm lg:text-base xl:text-xl">
>     {teamMember.name.split(' ')[0]}
>     <br />
>     {teamMember.name.split(' ')[1]}
>   </p>`
> ```
> 
> However, what should happen when someone enters more than two names?
> 
> image
> 
> Do you guys think we should stick to the mockups in this matter and if so, do you have any suggestions for longer names?
> 
> The second question is related to the hover state on mobile, as that is obviously not possible. I was thinking of replacing it by click, but I've looked into a couple of things and I don't know how to have that effect for mobile only. Suggestions would be appreciated.

1. I believe we can wrap the text into 2 lines independently of the number of names: https://stackoverflow.com/questions/11989546/wrap-a-text-within-only-two-lines-inside-div Please try it out and check if the result is what you want
2. I think the click makes sense

pedrosilva17:
> > I have two questions I'd like to discuss with you guys:
> > Based on these mockups, I tried to place first name and last name on different lines.
> > image
> > ```
> >   <p
> >     class="mx-auto text-center text-sm font-bold leading-tight text-gray-100 transition-all sm:text-sm md:text-sm lg:text-base xl:text-xl">
> >     {teamMember.name.split(' ')[0]}
> >     <br />
> >     {teamMember.name.split(' ')[1]}
> >   </p>`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However, what should happen when someone enters more than two names?
> > image
> > Do you guys think we should stick to the mockups in this matter and if so, do you have any suggestions for longer names?
> > The second question is related to the hover state on mobile, as that is obviously not possible. I was thinking of replacing it by click, but I've looked into a couple of things and I don't know how to have that effect for mobile only. Suggestions would be appreciated.
> 
> Came up with a working version for the second problem, but more concise suggestions would be appreciated

There are some issues with the solution you provided:
- It works for one hexagon, but not for multiple ones (which is going to happen 99% of the time). That's because you use `querySelector()` on elements with the `target` class, but with more than one hexagon displayed it's just going to select the first one.
- Even though only the first hexagon is being observed, clicking on it causes changes in the other hexagons: All of the team members' roles appear in their respective containers by tapping any hexagon. You query all of the containers with the `fullOpacityContainers` and `variableOpacityContainers` classes without verifying if they belong to the hexagon being clicked (unless you want to animate all of the hexagons upon clicking any of them, then it would make sense).
- I'm not sure I understood the `clickToggle` variable: Is the goal to animate each hexagon individually on click and collapse all of them when you close one?

The first thing to be done is to replace the `querySelector()` on the target classes by a `querySelectorAll()` and create a listener for each element.
If the goal is to make each hexagon independent by having its own state, open or closed (a simple binary array should take care of that), you have to apply your selectors to the current targeted hexagon and not the whole document, otherwise the second issue I mentioned above will keep happening. Just move that part of the code inside the logic of each listener and change `document.querySelectorAll()` to `target.querySelectorAll()` (for example).

If you need, I can show you my idea during the meeting, but you might want to work on it alone first after these suggestions. Up to you.

MRita443:
> > I have two questions I'd like to discuss with you guys:
> > Based on these mockups, I tried to place first name and last name on different lines.
> > image
> > ```
> >   <p
> >     class="mx-auto text-center text-sm font-bold leading-tight text-gray-100 transition-all sm:text-sm md:text-sm lg:text-base xl:text-xl">
> >     {teamMember.name.split(' ')[0]}
> >     <br />
> >     {teamMember.name.split(' ')[1]}
> >   </p>`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However, what should happen when someone enters more than two names?
> > image
> > Do you guys think we should stick to the mockups in this matter and if so, do you have any suggestions for longer names?
> > The second question is related to the hover state on mobile, as that is obviously not possible. I was thinking of replacing it by click, but I've looked into a couple of things and I don't know how to have that effect for mobile only. Suggestions would be appreciated.
> 
>     1. I believe we can wrap the text into 2 lines independently of the number of names: https://stackoverflow.com/questions/11989546/wrap-a-text-within-only-two-lines-inside-div Please try it out and check if the result is what you want
> 
>     2. I think the click makes sense

Thank you! When it comes to the first question, I tried this but it's not exactly what I am looking for. This limits the text to two lines, but in this case I would like to _force_ the text into two lines, even when it fits into one. I tried some things liking making the name a list-item and having a max width to force the name to split, only the latter works but I find it a bit iffy because it doesn't work for really short names and messes with really long words too.

MRita443:
> > > I have two questions I'd like to discuss with you guys:
> > > Based on these mockups, I tried to place first name and last name on different lines.
> > > image
> > > ```
> > >   <p
> > >     class="mx-auto text-center text-sm font-bold leading-tight text-gray-100 transition-all sm:text-sm md:text-sm lg:text-base xl:text-xl">
> > >     {teamMember.name.split(' ')[0]}
> > >     <br />
> > >     {teamMember.name.split(' ')[1]}
> > >   </p>`
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > However, what should happen when someone enters more than two names?
> > > image
> > > Do you guys think we should stick to the mockups in this matter and if so, do you have any suggestions for longer names?
> > > The second question is related to the hover state on mobile, as that is obviously not possible. I was thinking of replacing it by click, but I've looked into a couple of things and I don't know how to have that effect for mobile only. Suggestions would be appreciated.
> > 
> > 
> > Came up with a working version for the second problem, but more concise suggestions would be appreciated
> 
> There are some issues with the solution you provided:
> 
>     * It works for one hexagon, but not for multiple ones (which is going to happen 99% of the time). That's because you use `querySelector()` on elements with the `target` class, but with more than one hexagon displayed it's just going to select the first one.
> 
>     * Even though only the first hexagon is being observed, clicking on it causes changes in the other hexagons: All of the team members' roles appear in their respective containers by tapping any hexagon. You query all of the containers with the `fullOpacityContainers` and `variableOpacityContainers` classes without verifying if they belong to the hexagon being clicked (unless you want to animate all of the hexagons upon clicking any of them, then it would make sense).
> 
>     * I'm not sure I understood the `clickToggle` variable: Is the goal to animate each hexagon individually on click and collapse all of them when you close one?
> 
> 
> The first thing to be done is to replace the `querySelector()` on the target classes by a `querySelectorAll()` and create a listener for each element. If the goal is to make each hexagon independent by having its own state, open or closed (a simple binary array should take care of that), you have to apply your selectors to the current targeted hexagon and not the whole document, otherwise the second issue I mentioned above will keep happening. Just move that part of the code inside the logic of each listener and change `document.querySelectorAll()` to `target.querySelectorAll()` (for example).
> 
> If you need, I can show you my idea during the meeting, but you might want to work on it alone first after these suggestions. Up to you.

Thank you! I hadn't tested the component on a grid so I hadn't noticed these issues, I'm working on your suggestions. I used the `clickToggle` variable because, using only `touchstart` and `touchend`, the extra information would only show while the hexagon is pressed down, and I thought it would be a quality of life improvement to have them be tapped, maintain the information, and go back to their base state with another tap. I also chose to use these events because the goal is to have the click only work on mobile, and have hover for pc, and I am not aware of other ways to to that with different events.

pedrosilva17:
Good job, almost there! There's only one small issue with the `clickToggle` variable when multiple hexagons are present.

When you tap on a closed hexagon, it opens and increments the variable by 1. If you try to close that same hexagon, it behaves as intended, closing immediately and reverting the variable's value to 0. However, if you open a hexagon and then try to open another one in a row, it won't work, requiring you to tap it again.
This happens because of `clickToggle`: the variable is set to 0 initially, so when you try to open the first hexagon it passes the first check successfully (`!clickToggle`). When you do the same for the second hexagon, the variable is at 1, causing that check to fail and applying the transformations to close the hexagon, which obviously doesn't change it.

When you have multiple elements which can change independently on tap, you must keep track of the state of each element. You can do that by simply creating a binary array (0 means closed and 1 means open, initially filled with 0s) with size equal to the number of elements. If you open/close a button, simply change the number at the corresponding index in the array (change the `for...of` loop into a `for` loop with indices, making it easier to access the current index) to 0 or 1. Feel free to think of other solutions, as I might be missing an easier/cleaner way of doing this.

I'll review the rest of the PR again when I can, I'm typing this on my phone so I can't be as thorough as I would've liked.

MRita443:
> Good job, almost there! There's only one small issue with the `clickToggle` variable when multiple hexagons are present.
> 
> When you tap on a closed hexagon, it opens and increments the variable by 1. If you try to close that same hexagon, it behaves as intended, closing immediately and reverting the variable's value to 0. However, if you open a hexagon and then try to open another one in a row, it won't work, requiring you to tap it again. This happens because of `clickToggle`: the variable is set to 0 initially, so when you try to open the first hexagon it passes the first check successfully (`!clickToggle`). When you do the same for the second hexagon, the variable is at 1, causing that check to fail and applying the transformations to close the hexagon, which obviously doesn't change it.
> 
> When you have multiple elements which can change independently on tap, you must keep track of the state of each element. You can do that by simply creating a binary array (0 means closed and 1 means open, initially filled with 0s) with size equal to the number of elements. If you open/close a button, simply change the number at the corresponding index in the array (change the `for...of` loop into a `for` loop with indices, making it easier to access the current index) to 0 or 1. Feel free to think of other solutions, as I might be missing an easier/cleaner way of doing this.
> 
> I'll review the rest of the PR again when I can, I'm typing this on my phone so I can't be as thorough as I would've liked.

You're right, thanks! I tried to implement this using a data attribute for `state`, it seems to have worked well but I have come across an issue. Using a grid, when clicking on a hexagon it will be animated (opened and closed) the number of hexagons there are on the grid. Due to this, the animation is only visible when there is an odd number of hexagons on the grid.

I am not sure why this is happening, I used some prints to confirm the event is only triggered once, but it seems the event code is ran multiple times each event.

pedrosilva17:
I'm not sure how I didn't notice this earlier, but if we add an event listener for every hexagon each time a new one is created, of course it's going to add X amount of event listeners for the first hexagon, where X is the total amount of hexagons in the page. And with a page full of members... 😭

I can think of two ways of solving this issue right now: 
- If we're only having one grid per page, create a new component (a Team Member Hexagon grid) which displays any amount of team member hexagons, with the member array passed by parameter. Only then would we add the event listeners for each hexagon.
- Instead of using a class named `target`, select an id using the team member's own full name, or anything unique to him. Then add only one event listener to that hexagon, it should work correctly this time since every hexagon id is unique.

I don't know what I was thinking a few days ago, complete blunder on my part. Sorry!

MRita443:
> I'm not sure how I didn't notice this earlier, but if we add an event listener for every hexagon each time a new one is created, of course it's going to add X amount of event listeners for the first hexagon, where X is the total amount of hexagons in the page. And with a page full of members... sob
> 
> I can think of two ways of solving this issue right now:
> 
>     * If we're only having one grid per page, create a new component (a Team Member Hexagon grid) which displays any amount of team member hexagons, with the member array passed by parameter. Only then would we add the event listeners for each hexagon.
> 
>     * Instead of using a class named `target`, select an id using the team member's own full name, or anything unique to him. Then add only one event listener to that hexagon, it should work correctly this time since every hexagon id is unique.
> 
> 
> I don't know what I was thinking a few days ago, complete blunder on my part. Sorry!

No worries! I didn't notice until then because I only (conveniently) tested with an odd number of hexagons too. It's fixed now, I used email as an id.

Currently waiting on the design team to let me know if there should be a default profile pic (cc @NIAEFEUP/design-website), besides that and discussing the name splitting it should be done! Thanks for the suggestions

srissirs:
> > I'm not sure how I didn't notice this earlier, but if we add an event listener for every hexagon each time a new one is created, of course it's going to add X amount of event listeners for the first hexagon, where X is the total amount of hexagons in the page. And with a page full of members... sob
> > I can think of two ways of solving this issue right now:
> > ```
> > * If we're only having one grid per page, create a new component (a Team Member Hexagon grid) which displays any amount of team member hexagons, with the member array passed by parameter. Only then would we add the event listeners for each hexagon.
> > 
> > * Instead of using a class named `target`, select an id using the team member's own full name, or anything unique to him. Then add only one event listener to that hexagon, it should work correctly this time since every hexagon id is unique.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't know what I was thinking a few days ago, complete blunder on my part. Sorry!
> 
> No worries! I didn't notice until then because I only (conveniently) tested with an odd number of hexagons too. It's fixed now, I used email as an id.
> 
> Currently waiting on the design team to let me know if there should be a default profile pic (cc @NIAEFEUP/design-website), besides that and discussing the name splitting it should be done! Thanks for the suggestions

I have left an example of a default image in figma but here's the image
image

Really good work guys!

Sirze01 (approved):
It seems to be working as expected. 
I think we should merge this and start using it on the pages to catch any new strange behaviours and test it outside storybook.

Good work! :tada:

LuisDuarte1 (approved):
Great job! On my testing it looks perfect, congrats! :tada:

PR #139, opened 2023-04-05
Feature/contacts

### LINKED ISSUE #17: contacts: form page
Create the contacts page following the mockups and the screenshot below. Please implement both PC and mobile versions.

- [x] PC contacts page
- [x] Mobile contacts page

image

image

jlcrodrigues:
Closes #17

Created the contacts page for both PC and mobile.

# Screenshots

<p float="left">

</p>

# Review checklist

- [ ] Behavior is as expected


src/routes/contacts/+page.svelte
```diff
@@ -0,0 +1,74 @@
...
+    />
+    {:else}
+    <img
+        alt="Feup Buildings Outline"
+            src="/images/feup_buildings.svg"
+        class="align-center justify-self-center w-full"
+        style="min-width:400px; min-height:160px; object-fit: none; object-position: center;"
+
+    />
+    {/if}
```
pedrosilva17: - The inline style tags are not necessary and can be applied with tailwind (`object-none`, `object-center`, `min-w`and `min-h`), although with slightly different px values (for example, `h-52` is equal to `height:208px;` which I would say is close enough). If you really need exactly 200px, you can use `h-[200px]`, which is not as clean as the previous tailwind class. Moreover, some classes are not very useful from what I've seen, like the min-widths and min-heights (the latter of which can just be converted to height in this case, I think, as the difference is not noticeable). Again, if you feel like they should be preserved, you can pass an arbitrary value to the `min-w` and `min-h`classes using square brackets like I did above.
- This if/else statement can be replicated using a `<picture>` tag (read more about it [here](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#art_direction)). Avoids repeating common parts of the image, like the alt text and almost all classes, and doesn't need to use a variable that stores the width of the window, instead changing the source of the image based on a breakpoint. The tag is supported by [every modern browser](https://caniuse.com/picture) (post-2015), but we can discuss its usage in the meeting later today. For the classes that change depending on the window width, just use the tailwind `md:`breakpoint, which has the exact same px value as the `if` condition.

src/routes/contacts/+page.svelte
```diff
@@ -0,0 +1,74 @@
...
+    form > input {
+        width: 100%;
+    }
+
+    @media (min-width: 1050px) {
+        #location {
+            width: 1039px;
+        }
+    }
+</style>
```
pedrosilva17: These can be achieved with tailwind as well.
- Use `w-full` in each input's class
- Use `lg:w-[1039px]` inside the span's class (the breakpoint targets devices with 1024px of width or greater and sets the width of the span to 1039px in those cases).

Keep in mind the arbitrary value attribution with square brackets is reserved for situations where there are no native similar values, which is the case here.

src/routes/contacts/+page.svelte
```diff
@@ -0,0 +1,74 @@
...
+    <div class="md:grid grid-cols-2 mx-10">
+        <form class="justify-self-end min-w-[70%]">
+            <label class="m-1 text-white font-source-code" for="email">// Email</label><br>
+            <input class="rounded-lg p-1 text-primary placeholder-primary" type="email" id="email" placeholder="xxxxxxxxxxxxx@gmail.com"><br>
+            <label class="m-1 text-white font-source-code" for="name">// Nome</label><br>
+            <input class="rounded-lg p-1 text-primary placeholder-primary" type="text" id="name"><br>
+            <label class="m-1 text-white font-source-code" for="subject">// Assunto</label><br>
+            <input class="rounded-lg p-1 text-primary placeholder-primary" type="text" id="subject"><br>
+            <label class="m-1 text-white font-source-code" for="message">// Mensagem</label><br>
+            <textarea class="rounded-lg p-1 text-primary placeholder-primary" id="message"></textarea><br>
```
pedrosilva17: - Maybe each label/input pair could be converted into a component? I imagine we could potentially have inputs like these in other places of the website, plus it improves readability and reduces code repetition. The parameters to be passed could be the `for`, `type`and `placeholder` labels.
- The font is actually named `source_code`and not `source-code`, I changed it in my commit to avoid putting brackets around the name in tailwind config file. In hindsight, that might not have been the best decision, as it looks strange in a tailwind class. Maybe it could be reverted and you don't need to change the name here, but just know that the font is not being rendered properly with this code. We'll discuss that in the meeting.

pedrosilva17 (commented):
Great work! I have some suggestions that might improve the code a little bit, let me know if you agree with them.

srissirs (dismissed):
It's looking good! Though I think there should be more space between the fields. The titles are a bit too close to the field directly above. Also, the graph in the mobile version does not seem to be centered. Could you fix that too?

jlcrodrigues:
@srissirs @pedrosilva17 Thank you for the feedback. I think I fixed the problems mentioned. Please take another look.

src/routes/contacts/Graph.svelte
```diff
@@ -0,0 +1,82 @@
...
+  ];
+  let innerWidth = 0;
+  let coefficient = 8;
+  const iconSize = 8;
+
+  const socials = [
+    { url: 'https://www.instagram.com/niaefeup/', icon: FaBrandsInstagram },
+    { url: 'https://github.com/NIAEFEUP', icon: FaBrandsGithub },
+    { url: 'https://www.facebook.com/NIAEFEUP', icon: FaBrandsFacebook },
+    { url: 'ni@aefeup.pt', icon: IoMail },
```
BrunoRosendo: the email should be copied to the clipboard, that doesn't seem to be working

src/routes/contacts/+page.svelte
```diff
@@ -0,0 +1,47 @@
+<script>
+  import LabelInput from '$lib/component/LabelInput.svelte';
+  import Graph from './Graph.svelte';
+  import Icon from 'svelte-icons-pack/Icon.svelte';
+  import BiMap from 'svelte-icons-pack/bi/BiMap';
+</script>
+
+<main class="-10 my-20 flex flex-col justify-center ">
```
BrunoRosendo:
```suggestion
<main class="-10 my-20 flex flex-col justify-center">
```
Linter doesn't catch this? 🤔

src/routes/contacts/+page.svelte
```diff
@@ -0,0 +1,47 @@
+<script>
+  import LabelInput from '$lib/component/LabelInput.svelte';
+  import Graph from './Graph.svelte';
+  import Icon from 'svelte-icons-pack/Icon.svelte';
+  import BiMap from 'svelte-icons-pack/bi/BiMap';
```
BrunoRosendo: please use the `Icons`  pattern we used for the footer. Add the export in the `Icons.ts` and use it in all icons' sources

src/routes/contacts/Graph.svelte
```diff
@@ -0,0 +1,82 @@
+<script>
+  import Icon from '$lib/component/Icon.svelte';
+  import FaBrandsInstagram from 'svelte-icons-pack/fa/FaBrandsInstagram';
+  import FaBrandsTwitter from 'svelte-icons-pack/fa/FaBrandsTwitter';
+  import FaBrandsFacebook from 'svelte-icons-pack/fa/FaBrandsFacebook';
+  import FaBrandsGithub from 'svelte-icons-pack/fa/FaBrandsGithub';
+  import IoMail from 'svelte-icons-pack/io/IoMail';
+  import FaBrandsLinkedin from 'svelte-icons-pack/fa/FaBrandsLinkedin';
```
BrunoRosendo: Same as above

src/routes/contacts/Graph.svelte
```diff
@@ -0,0 +1,82 @@
...
+            href={socials[index].url}
+          />
+        </div>
+      {:else}
+        <div
+          class="flex h-full w-full items-center justify-center rounded bg-white bg-opacity-30"
+          on:click={() => copyToClipboard(socials[index].url)}
+          on:keydown
+        >
+          <Icon src={socials[index].icon} color="white" size={iconSize.toString()} />
```
BrunoRosendo: I think size is expecting something like "8px", not "8". Please check the behaviour on this

src/routes/contacts/Graph.svelte
```diff
@@ -0,0 +1,82 @@
...
+
+  <!-- Draw graph nodes. -->
+  {#each coords as coord, index}
+    <foreignObject
+      x={coord[0] * coefficient - (iconSize + 2) / 2}
+      y={coord[1] * coefficient - (iconSize + 2) / 2}
+      width={iconSize + 2}
+      height={iconSize + 2}
+    >
+      {#if socials[index].url.slice(0, 5) == 'https'}
```
BrunoRosendo:
```suggestion
      {#if socials[index].url.startsWith('https')}
```
More appropriate

src/routes/contacts/Graph.svelte
```diff
@@ -0,0 +1,82 @@
...
+            color="white"
+            size={iconSize.toString()}
+            href={socials[index].url}
+          />
+        </div>
+      {:else}
+        <div
+          class="flex h-full w-full items-center justify-center rounded bg-white bg-opacity-30"
+          on:click={() => copyToClipboard(socials[index].url)}
+          on:keydown
```
BrunoRosendo: why is `on:keydown` needed?

BrunoRosendo (changes requested):
Good job so far! I have a few suggestions

src/routes/contacts/+page.svelte
BrunoRosendo: layout already has a `<main>`, maybe this should be something else

src/routes/contacts/Graph.svelte
jlcrodrigues: visible, non-interactive elements with an on:click event must be accompanied by an on:keydown, on:keyup, or on:keypress event

src/routes/contacts/Graph.svelte
jlcrodrigues: It's not clickable because the logo image is overlapping it (#146). Otherwise, it is working.

src/routes/contacts/Graph.svelte
```diff
@@ -0,0 +1,77 @@
...
+      y={coord[1] * coefficient - (iconSize + 2) / 2}
+      width={iconSize + 2}
+      height={iconSize + 2}
+    >
+      {#if socials[index].url.slice(0, 5) == 'https'}
+        <div class="flex h-full w-full items-center justify-center rounded bg-white bg-opacity-30">
+          <Icon
+            src={socials[index].icon}
+            color="white"
+            size={iconSize.toString()}
```
BrunoRosendo:
```suggestion
            size="{iconSize.toString()}px"
```

src/routes/contacts/Graph.svelte
```diff
@@ -0,0 +1,77 @@
...
+
+  <!-- Draw graph nodes. -->
+  {#each coords as coord, index}
+    <foreignObject
+      x={coord[0] * coefficient - (iconSize + 2) / 2}
+      y={coord[1] * coefficient - (iconSize + 2) / 2}
+      width={iconSize + 2}
+      height={iconSize + 2}
+    >
+      {#if socials[index].url.slice(0, 5) == 'https'}
```
BrunoRosendo: This didn't change, did you try `startsWith`?

src/routes/contacts/+page.svelte
```diff
@@ -0,0 +1,47 @@
...
+        type="submit">Enviar</button
+      >
+    </form>
+    <div class="m-5 flex w-full justify-center md:my-0">
+      <Graph />
+    </div>
+  </div>
+  <picture>
+    <source media="(max-width: 768px)" srcset="/images/feup_buildings.svg" />
+    <source media="(min-width: 769px)" srcset="/images/feup_buildings_md.svg" />
```
BrunoRosendo: why doesn't it use the same pixels?

BrunoRosendo (changes requested):
just small stuff

static/images/outline_white.png
LuisDuarte1, line 1: Any reason why you've scaled the hexagon outline?

LuisDuarte1 (commented):
Overall it looks great, great job :rocket:, there's some issues with z-index and the hexagon outline, which I believe that are already fixed on `develop`? So try to merge and see if that's fixed ^^. Just have some minor concern below, but great job otherwise :tada:

BrunoRosendo (approved):
I'm approving since the feature is looking great! There are just a few side things missing:
- Fix lint
- Create storybook stories and tests (I'll leave this as the team's discretion, since you may want to save time on this)

package-lock.json
```diff
@@ -19878,11 +19776,6 @@
       "integrity": "sha512-3wdGidZyq5PB084XLES5TpOSRA3wjXAlIWMhum2kRcv/41Sn2emQ0dycQW4uZXLejwKvg6EsvbdlVL+FYEct7A==",
       "dev": true
     },
-    "node_modules/tabbable": {
-      "version": "6.2.0",
-      "resolved": "https://registry.npmjs.org/tabbable/-/tabbable-6.2.0.tgz",
-      "integrity": "sha512-Cat63mxsVJlzYvN51JmVXIgNoUokrIaT2zLclCXjRd8boZ0004U4KCs/sToJ75C6sdlByWxpYnb5Boif1VSFew=="
-    },
```
MRita443: Were these removed dependencies not being used?

src/lib/components/icons/LabelInput.svelte
```diff
@@ -0,0 +1,23 @@
+<script>
```
MRita443: Because this component is reusable, it should be placed in `src/lib/components`. Also, please follow the lower caps file name and add this component to storybook!

src/routes/(app)/contacts/Graph.svelte
```diff
@@ -0,0 +1,81 @@
+<script lang="ts">
```
MRita443: This should be in `src/routes/contacts/_components` for consistency, and the file name lowercaps too

MRita443 (changes requested):
Looking good overall, a couple file placement and file name consistency changes since the restructuring, and please add the new component to storybook. Otherwise (this might be a bit nit picky) but I noticed when scaling the page at one point I shrink the page and the forms gets bigger which looks a bit odd. Is this easily fixable? If not I say we keep it this way

[Screencast from 03-08-2024 12:44:24.webm](https://github.com/user-attachments/assets/171549b1-a88f-4b6b-9cb9-7b52f11bb062)

package-lock.json
rubuy-74: I didn't change any of the dependencies, the file was generated like this.

src/lib/components/icons/label-input.svelte
```diff
@@ -0,0 +1,23 @@
...
+    rows="4"
+  /><br />
+{:else}
+  <input
+    class="mb-2 w-full rounded-lg p-2 text-primary placeholder-primary"
+    {type}
+    {id}
+    {placeholder}
+  /><br />
+{/if}
```
MRita443: I think it'd be beneficial to put this component on storybook

src/routes/(app)/contacts/graph.svelte
```diff
@@ -0,0 +1,81 @@
...
+          role="button"
+          tabindex="0"
+          aria-label="copy-mail"
+        >
+          <Icon src={socials[index].icon} color="white" size="{iconSize.toString()}px" />
+        </div>
+      {/if}
+    </foreignObject>
+  {/each}
+</svg>
```
MRita443: Although less reusable it'd be nice to have this on storybook too, for example to test responsivity easily :) And please place this in `src/routes/(app)/contacts/_components` for consistency

MRita443 (changes requested):
Seems ready but just some small comments!

MRita443:
@rubuy-74 Storybook tests are failing due to some accessibility checks, pls fix them! To see what they are run storybook and check the accessibility tab at the bottom of the page

MRita443 (changes requested):
Everything from my previous review looks great, two small things. The inputs appear as black to me, probably because of my browsers theme.
image
Second thing, the `LabelInput` component, when a text area, still gives the option for a placeholder but no placeholder is visible (try it out on storybook)

MRita443:
@srissirs Please re-review!

MRita443 (approved):
Great job! Having to add the `fixed` back into the `navbar` creates the problem of making page content scroll behind it I opened an [issue](https://github.com/NIAEFEUP/website-niaefeup-frontend/issues/291) for that, but this page looks great!

PR #163, opened 2023-04-28
Events listing page

### LINKED ISSUE #14: events: basic listing page
Create a first version for the `Events` page. Since the complete version is a big addition to the site, we can start by implementing a page with all the UI components (using what's being created in #1 ), and information retrieval but little to no user interaction. The included features would be:
- [ ] List of events, making use of the work in #1 
- [ ] Auxiliary UI components, such as title, search bar, filters and timeline (look at the mockups and screenshot below). Note that these should be reusable and added to storybook
- [ ] Mobile view for the previous points
- [ ] Retrieve events list from the backend

image

Image

### LINKED ISSUE #146: Fix background image overlapping other components
Currently, the background image of the layout is positioned above other content, resulting in non-interactive buttons and components. For example, you can easily see this with the mobile navbar and footer links for social media.

This seems to be a problem related to z-index in the layout and its body

DoStini:
Closes #14

# Screenshots

Generalized TextField that allows pre and after icons that might act as buttons (see IconOptions)

image

# Review checklist

- [ ] Behavior is as expected


BrunoRosendo:
image
please fix, related to z-index

package.json
```diff
@@ -49,6 +49,7 @@
   },
   "type": "module",
   "dependencies": {
-    "@sveltejs/adapter-auto": "^2.0.0"
+    "@sveltejs/adapter-auto": "^2.0.0",
+    "svelte-select": "^5.6.0"
```
BrunoRosendo, line 66: move to dev deps is compiled

BrunoRosendo:
open issue in svelte-select repo because of dropdown not closing

limwa:
Regarding the select component, React has a very popular library for customizable components called Radix UI (https://www.radix-ui.com/). Although the original project doesn't support Svelte yet, we can look for some "ports", like https://www.radix-svelte.com/ (replaced by https://www.melt-ui.com/). These two last libraries are not production ready, but I'll keep looking for some libraries similar to Radix UI for Svelte

limwa:
It appears the Svelte ecosystem in this area is still a bit underdeveloped... Going with custom components made by ourselves seems like the safest bet.

PR #164, opened 2023-04-30
Storybook interactions and unit tests

### LINKED ISSUE #8: tests: configure and create initial example tests

BrunoRosendo:
closes #8 

This PR adds interactions tests using storybook to the current stories and creates some unit tests for lib functions using vitest.

npm test is now a simple pipeline with 2 steps: vitest runner and storybook's test-runner

Please give suggestions on additional things we might want to test

# Review checklist

- [ ] Behavior is as expected


BrunoRosendo:
This still needs some CI changes in the github workflows

BrunoRosendo:
Almost ready, I just want to setup codecov

src/routes/api/[...endpoint]/+server.ts
```diff
@@ -1,10 +1,6 @@
 import type { RequestHandler } from '@sveltejs/kit';
 import { fetchWithAuth } from './proxy';
-
-function endpoint(url: URL) {
-  const pathName = url.pathname.replace('/api', '');
-  return pathName.startsWith('/') ? pathName.substring(1) : pathName;
-}
+import { endpoint } from '../../../lib/api/proxy';
```
Comment (LuisDuarte1) [inline src/routes/api/[...endpoint]/+server.ts]: Maybe avoid using relative imports, if possible.

src/routes/layout.stories.ts
```diff
@@ -10,4 +12,62 @@ export default {
...
 
-export const Default = {};
+export const Default = {
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+
+    await step('Copy email and check notification', async () => {
+      await userEvent.click(await canvas.findByTestId('email-icon'));
+      const notification = canvas.queryByText('O email foi copiado para o teu clipboard :)');
+      await expect(notification).toBeTruthy();
```
LuisDuarte1: just a suggestion and maybe this comes to personal preference, make a this string a global const variable, to make changes easier in the future, since you use it multiple times. Other thing I don't know if multiple languages are to be supported in the future and how this will affect tests.

LuisDuarte1 (changes requested):
Overall the introduction of testing seems lgtm, just have some suggestions and questions below.

LuisDuarte1:
Also other thing [a11y integration](https://storybook.js.org/docs/react/writing-tests/accessibility-testing) seems interesting for fixing accessibility issues (eg.: making a link not with a anchor element but with a div with a onClick function) wdyt? maybe this should be another pr idk

src/routes/layout.stories.ts
BrunoRosendo: I agree. With translations, we'd likely have a map with different languages instead of constants. Imo that's a moderate refactor for the future

pedrosilva17 (commented):
I agree with Luís's assessment. Regarding the multiple language support, I think we should have English alongside Portuguese, at least. It's a refactor to consider in the future, like you said.

src/lib/layout/Sidebar.stories.ts
```diff
@@ -13,5 +15,20 @@ export const MobileSidebar = {
...
       defaultViewport: 'mobile2'
     }
+  },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+
+    await step('Open sidebar', async () => {
+      await userEvent.click(await canvas.findByRole('button'));
+      const menuOption = canvas.queryByText('Início');
+      await expect(menuOption).toBeTruthy();
```
pedrosilva17: Maybe this would also need to be changed when we implement multiple language support.

src/routes/api/[...endpoint]/+server.ts
Comment (BrunoRosendo) [reply src/routes/api/[...endpoint]/+server.ts]: I'll check if there's any rule we can add on the linter for that

Comment (BrunoRosendo) [reply src/routes/api/[...endpoint]/+server.ts]: I didn't find a rule for that, unfortunately :/

BrunoRosendo:
> Also other thing [a11y integration](https://storybook.js.org/docs/react/writing-tests/accessibility-testing) seems interesting for fixing accessibility issues (eg.: making a link not with a anchor element but with a div with a onClick function) wdyt? maybe this should be another pr idk

I added the plugin and configured test-runner with it. However, I think it is pretty strict. I had a situation (member button and footer) where I had to deactivate the colour-contrast rule.
I think we can either work like this or perhaps make our own a11y configuration

BrunoRosendo:
> > Also other thing [a11y integration](https://storybook.js.org/docs/react/writing-tests/accessibility-testing) seems interesting for fixing accessibility issues (eg.: making a link not with a anchor element but with a div with a onClick function) wdyt? maybe this should be another pr idk
> 
> I added the plugin and configured test-runner with it. However, I think it is pretty strict. I had a situation (member button and footer) where I had to deactivate the colour-contrast rule. I think we can either work like this or perhaps make our own a11y configuration

I solved this by configuring a11y to review contrast changes instead of failing on the test runner. It still appears on the UI but the CI passes

pedrosilva17:
About the relative imports: If I understood correctly, this is the rule we're looking for: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-relative-parent-imports.md
Can we add this plugin to our project?

BrunoRosendo:
> About the relative imports: If I understood correctly, this is the rule we're looking for: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-relative-parent-imports.md Can we add this plugin to our project?

It is, thanks! I will try to install the plugin

BrunoRosendo:
@pedrosilva17 @LuisDuarte1 I added the eslint import plugin and a few rules, please check them out and see if you agree

pedrosilva17:
LGTM. A bit strict, but it will make it easier to enforce consistent imports across the code.

LuisDuarte1 (approved):
LGTM :rocket:

pedrosilva17 (approved):
Forgot to accept the changes, sorry

PR #214, opened 2023-07-12
Login page

### LINKED ISSUE #140: auth: implement login page
Create initial version of login page, including:

- [ ] Email and password input fields, such as depicted below
- [ ] Reusable show/hide password component (should be added to storybook)
- [ ] User feedback when login is unsuccessful
- [ ] Mobile view

image

MRita443:
Closes #140

Initial implementation of member login page

# Screenshots

image

# Review checklist

- [ ] Behavior is as expected


LuisDuarte1 (commented):
Good progress overall 🚀, besides responsiveness and the color issue in tailwind which you are working on, the scaling looks weird on a display with lower vertical resolution (which is the first image in this case, and the case with my laptop 😄 ) maybe the login should scale down accordingly to make the experience more consistent

tailwind.config.cjs
```diff
@@ -12,7 +15,10 @@ module.exports = {
         primary: primaryColor,
         secondary: secondaryColor,
         tertiary: tertiaryColor,
-        tertiary60: tertiary60Color
+        tertiary60: tertiary60Color,
+        quaternary: quaternaryColor,
+        quaternary54: quaternary54Color,
+        lightRed: lightRedColor
```
LuisDuarte1: Maybe look into [Background Opacity](https://v2.tailwindcss.com/docs/background-opacity) but it only works for background or even [Customizing colors](https://tailwindcss.com/docs/customizing-colors#using-css-variables) that suports an alpha placeholder in order to remove some repeated colors

LuisDuarte1 (changes requested):
Good job making this work on all desktops :partying_face: but I have some concerns about the colors that we talked about in the retrospective

src/lib/component/Icon.svelte
```diff
@@ -13,5 +13,5 @@
     <Icon {src} {color} {size} />
   </a>
 {:else}
-  <Icon {src} {color} {size} />
+  <Icon {src} {size} color="currentcolor" className={$$props.color} />
```
BrunoRosendo: Why do you use `$$props.color`? Isn't simply `color` the same? Maybe I'm missing something here

src/lib/component/VariableVisibilityInput.svelte
```diff
@@ -0,0 +1,19 @@
...
+  import Icon from './Icon.svelte';
+  import Icons from './Icons';
+
+  let visible = false;
+  $: type = visible ? 'text' : 'password';
+  $: icon = visible ? Icons.Hidden : Icons.Visible;
+</script>
+
+<div class="grid">
+  <input {type} {...$$restProps} class="col-start-1 col-end-3 row-start-1 {$$props.class}" />
```
BrunoRosendo: I didn't know restProps, what it the goal here? I searched a little bit and it doesn't seem to be recommended https://github.com/sveltejs/svelte/issues/8249

BrunoRosendo (changes requested):
I'm not sure why but the small icon my password manager adds really screws up the password container. This is a bit concerning since I've never seen this before and password managers are pretty popular. Can you investigate?

image

src/lib/component/Icon.svelte
BrunoRosendo: I'd also appreciate an explanation about className, it just reminds me of react xd

BrunoRosendo:
Also, it'd be nice to have this in storybook too

src/lib/component/VariableVisibilityInput.svelte
MRita443: My goal here was to make the component reusable in different scenarios, but maintain the same icon placement. For that I had to have these specific classes, but still wanted to allow for more customization, hence the `$$props.class`. The `$$restProps` was to collect all the other attributes like names, placeholders, values etc. I decided to use this because inputs have lots of possible attributes, but I guess I could try to make the most common ones into optional properties maybe?

src/lib/component/Icon.svelte
MRita443: > I'd also appreciate an explanation about className, it just reminds me of react xd

image

It's just how the icon pack declares it ahah, otherwise I get a `class does not exist in type` error.
Just tried it myself, I think `class` is a reserved word

MRita443: > Why do you use `$$props.color`? Isn't simply `color` the same? Maybe I'm missing something here

Yup! Will change it. Also thinking of instead of this line simply adding a `className` property, for when the color trying to be used is not recognized by the `color` property (eg. custom names, different opacity notation). This way the color wouldn't have to be something like `fill-rose-950` and could actually just be a hex or color xd. What do you think?

MRita443:
> I'm not sure why but the small icon my password manager adds really screws up the password container. This is a bit concerning since I've never seen this before and password managers are pretty popular. Can you investigate?
> 
> image

I tried replicating this placing small icons on top of the input but I couldn't. Could you maybe inspect the page and share the code, or let me know the extension name?

src/lib/component/Icon.svelte
BrunoRosendo: @MRita443 I think `className` is specific to the `Icon` library we use. I prefer to use `$$props.class` and map it to `className` in `Icon.svelte`, so we have the same syntax for every component. But I'm willing to hear more opinions

src/lib/component/VariableVisibilityInput.svelte
BrunoRosendo: After searching a bit more about restProps, I believe this implementation is appropriate. Svelte has since updates the docs on this:
```
$$props references all props that are passed to a component, including ones that are not declared with export. Using $$props will not perform as well as references to a specific prop because changes to any prop will cause Svelte to recheck all usages of $$props. But it can be useful in some cases – for example, when you don't know at compile time what props might be passed to a component.

<Widget {...$$props} />

$$restProps contains only the props which are not declared with export. It can be used to pass down other unknown attributes to an element in a component. It shares the same performance characteristics compared to specific property access as $$props.

<input {...$$restProps} />
```

This means that specific prop access could be slower (such as $$props.class). I found an alternative but I don't find it visually appealing. Let me know what you think:

```js
export {_class as class};
let _class = "";

<input {type} {...$$restProps} class="{_class}" />
```

src/routes/login/page.stories.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default {
+  title: 'Pages/Login',
+  component: Page,
+  parameters: {
+    layout: 'fullscreen',
+    backgrounds: { default: 'clear' }
+  },
+  decorators: [() => Layout, () => LayoutDecorator]
+};
```
BrunoRosendo: It'd be nice to test the visibility input

BrunoRosendo (changes requested):

LuisDuarte1 (dismissed):
Nice progress, it's not 100% responsive yet though, as you can see on my phone, the login is not properly centered and overlapping the hamburger menu: 


Also it seems that you have an issue with importing layout from `login/page.stories.ts` that causes storybook to fail 😅

MRita443:
@BrunoRosendo @LuisDuarte1 pls re-review :)

src/routes/(app)/_components/layout/sidebar.svelte
```diff
@@ -2,16 +2,16 @@
   import SidebarItems from './sidebar-items';
   import NavItem from './nav-item.svelte';
   import BackgroundHexagon from './background-hexagon.svelte';
-  import Icon from '$lib/components/icons/icon.svelte';
-  import Icons from '$lib/components/icons/icons';
+  import Icon from '@/lib/components/icons/icon.svelte';
```
BrunoRosendo: curious, why @?

src/routes/(app)/login/+page.svelte
```diff
@@ -18,14 +20,37 @@
...
-  <br />
-  <input type="password" name="password" placeholder="password" bind:value={password} />
-  <br />
-  <button type="submit">Login</button>
-  <br />
-  <p>{message}</p>
-  <br />
-  <button><a href="/profile">Profile</a></button>
-</form>
+<!-- Colocar a página toda scrollable incluindo footer, ver container queries -->
```
BrunoRosendo:
```suggestion
```

src/routes/api/[...endpoint]/proxy.ts
```diff
@@ -12,7 +13,7 @@ async function _fetchApi(
   headers ??= new Headers();
   headers.append('Content-Type', 'application/json');
   headers.append('Accept', 'application/json');
-  if (window?.location?.origin) {
+  if (browser && window?.location?.origin) {
```
Comment (BrunoRosendo) [inline src/routes/api/[...endpoint]/proxy.ts:16]: when is the condition needed?

src/routes/(app)/login/+page.svelte
BrunoRosendo: is this still a problem? I don't think so

BrunoRosendo (changes requested):
really minor stuff

The page looks good IMO and the PR is pretty much ready for merge

src/routes/(app)/login/+page.svelte
MRita443: It seems like it is a problem on really short screens, but a general one at that because it happens not only on this page.
image
Do you think this is worth concerning about? I might open a new issue on it

src/routes/(app)/_components/layout/sidebar.svelte
MRita443: I really didn't think much of it and just used what the editor recommended me ahah but yeah in this case gonna switch back to `$lib`

src/routes/api/[...endpoint]/proxy.ts
Comment (MRita443) [reply src/routes/api/[...endpoint]/proxy.ts:16]: I was getting an error without this when submitting the forms:
`ReferenceError: window is not defined
    at _fetchApi (/home/rita/Documents/ni_site/website-niaefeup-frontend/src/routes/api/[...endpoint]/proxy.ts:16:3)
    at Module.fetchWithAuth (/home/rita/Documents/ni_site/website-niaefeup-frontend/src/routes/api/[...endpoint]/proxy.ts:44:12)
    at dispatchToBackend (/home/rita/Documents/ni_site/website-niaefeup-frontend/src/routes/api/[...endpoint]/+server.ts:12:10)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Module.render_endpoint (/home/rita/Documents/ni_site/website-niaefeup-frontend/node_modules/@sveltejs/kit/src/runtime/server/endpoint.js:43:18)
    at async resolve (/home/rita/Documents/ni_site/website-niaefeup-frontend/node_modules/@sveltejs/kit/src/runtime/server/respond.js:446:17)
    at async Object.handle (/home/rita/Documents/ni_site/website-niaefeup-frontend/src/hooks.server.ts:27:20)
    at async Module.respond (/home/rita/Documents/ni_site/website-niaefeup-frontend/node_modules/@sveltejs/kit/src/runtime/server/respond.js:322:20)
    at async file:///home/rita/Documents/ni_site/website-niaefeup-frontend/node_modules/@sveltejs/kit/src/exports/vite/dev/index.js:524:22`

src/routes/(app)/login/+page.svelte
BrunoRosendo: Ah I see! Open an issue, we can think about a fix later on

src/routes/api/[...endpoint]/proxy.ts
Comment (BrunoRosendo) [reply src/routes/api/[...endpoint]/proxy.ts:16]: An article about a similar error: https://dev.to/vvo/how-to-solve-window-is-not-defined-errors-in-react-and-next-js-5f97
I think this is a fine fix

BrunoRosendo (approved):

src/routes/(app)/login/+page.svelte
MRita443: I think this is fixed by #278, I thought this was a problem just with the mobile sidebar and didn't notice this case but it's fixed too

MRita443:
@LuisDuarte1 pls re-review

PR #215, opened 2023-07-14
Fix/proxy cors

### LINKED ISSUE #158: api: Fix proxy not using CORS for backend requests
We faced a problem in https://github.com/NIAEFEUP/website-niaefeup-backend/issues/58 when trying to test CORS that were configured on the backend. The configuration works well except when testing with frontend.

The problem likely resides in the proxy we're using to communicate authenticated requests with the backend. Since it sends requests from the server instead of the browser, CORS' headers are not used, CORS is only the default behaviour when requesting from the browser.

To solve this, we can add the headers manually in the proxy's requests. Let me know if you have additional solutions.

bdmendes:
Closes #158

# Review checklist

- [ ] Behavior is as expected

### DIFF
--- a/src/routes/api/[...endpoint]/+server.ts
+++ b/src/routes/api/[...endpoint]/+server.ts
@@ -11,3 +11,4 @@ export const GET: RequestHandler = dispatchToBackend;
 export const POST: RequestHandler = dispatchToBackend;
 export const PUT: RequestHandler = dispatchToBackend;
 export const DELETE: RequestHandler = dispatchToBackend;
+export const PATCH: RequestHandler = dispatchToBackend;
--- a/src/routes/api/[...endpoint]/+server.ts
+++ b/src/routes/api/[...endpoint]/+server.ts
@@ -12,3 +12,5 @@ export const POST: RequestHandler = dispatchToBackend;
 export const PUT: RequestHandler = dispatchToBackend;
 export const DELETE: RequestHandler = dispatchToBackend;
 export const PATCH: RequestHandler = dispatchToBackend;
+export const OPTIONS: RequestHandler = dispatchToBackend;
+export const HEAD: RequestHandler = dispatchToBackend;
--- a/src/routes/api/[...endpoint]/+server.ts
+++ b/src/routes/api/[...endpoint]/+server.ts
@@ -3,7 +3,12 @@ import { fetchWithAuth } from './proxy';
 import { endpoint } from '$lib/api/proxy';
 
 const dispatchToBackend: RequestHandler = async (event) => {
-  const body = event.request.method === 'GET' ? undefined : await event.request.text();
+  const body =
+    event.request.method === 'GET' ||
+    event.request.method === 'OPTIONS' ||
+    event.request.method === 'HEAD'
+      ? undefined
+      : await event.request.text();
   return fetchWithAuth(event.cookies, endpoint(event.url), event.request.method, undefined, body);
 };
 
--- a/src/routes/api/[...endpoint]/proxy.ts
+++ b/src/routes/api/[...endpoint]/proxy.ts
@@ -11,6 +11,10 @@ async function _fetchApi(
   const url = new URL(relativeUrl, PUBLIC_API_URL);
   headers ??= new Headers();
   headers.append('Content-Type', 'application/json');
+  headers.append('Accept', 'application/json');
+  if (window?.location?.origin) {
+    headers.append('Origin', window.location.origin);
+  }
   return fetch(url, { method: method, body, headers });
 }
 


src/routes/api/[...endpoint]/+server.ts
```diff
@@ -11,3 +11,6 @@ export const GET: RequestHandler = dispatchToBackend;
 export const POST: RequestHandler = dispatchToBackend;
 export const PUT: RequestHandler = dispatchToBackend;
 export const DELETE: RequestHandler = dispatchToBackend;
+export const PATCH: RequestHandler = dispatchToBackend;
+export const OPTIONS: RequestHandler = dispatchToBackend;
+export const HEAD: RequestHandler = dispatchToBackend;
```
Comment (limwa) [inline src/routes/api/[...endpoint]/+server.ts:21]: Is adding `CONNECT` and `TRACE` as well worth it?

limwa (approved):
LGTM :rocket:
Left a comment, but it's just a question and not really a request for changes!

limwa:
Also, I know that it is out of the scope of this PR but won't this line (https://github.com/NIAEFEUP/website-niaefeup-frontend/pull/215/files#diff-4bcda89884a979801f8a542290893a21e5213fc48218ec305b5434ef11e70ddaR6) cause an error when the request body is, for instance, an image?

src/routes/api/[...endpoint]/proxy.ts
```diff
@@ -11,6 +11,10 @@ async function _fetchApi(
   const url = new URL(relativeUrl, PUBLIC_API_URL);
   headers ??= new Headers();
   headers.append('Content-Type', 'application/json');
+  headers.append('Accept', 'application/json');
+  if (window?.location?.origin) {
+    headers.append('Origin', window.location.origin);
+  }
```
Comment (limwa) [inline src/routes/api/[...endpoint]/proxy.ts:17]: I think the `Origin` header is automatically added by browsers when `fetch` is called, so it shouldn't be necessary.

limwa (changes requested):
Sorry, I just noticed that `HEAD` and `OPTIONS` requests don't have a body. Line 6 of `+server.ts` should be changed accordingly. I also left another suggestion so let me know what you think.

src/routes/api/[...endpoint]/proxy.ts
Comment (bdmendes) [reply src/routes/api/[...endpoint]/proxy.ts:17]: It is, it's just a way to make sure it is there. I'm not sure how could we have a cors problem otherwise.

src/routes/api/[...endpoint]/+server.ts
Comment (bdmendes) [reply src/routes/api/[...endpoint]/+server.ts:21]: Do not think so, they are not the type of requests we would do to the server via the frontend

bdmendes:
> Also, I know that it is out of the scope of this PR but won't this line (https://github.com/NIAEFEUP/website-niaefeup-frontend/pull/215/files#diff-4bcda89884a979801f8a542290893a21e5213fc48218ec305b5434ef11e70ddaR6) cause an error when the request body is, for instance, an image?

I am not sure that it would cause an error (did a quick search and [the docs](https://developer.mozilla.org/en-US/docs/Web/API/Request/text) do not appear to mention that), maybe it will simply be a utf-8 string representation of the bytes.
I do agree that that may not be the best way to handle those cases, feel free to open an issue about that.

src/routes/api/[...endpoint]/proxy.ts
Comment (limwa) [reply src/routes/api/[...endpoint]/proxy.ts:17]: I think the cors problem was because `OPTIONS` requests (used by CORS) weren't being proxied to the backend. I'll approve either way.

limwa (approved):
LGTM :rocket:

limwa:
> > Also, I know that it is out of the scope of this PR but won't this line (https://github.com/NIAEFEUP/website-niaefeup-frontend/pull/215/files#diff-4bcda89884a979801f8a542290893a21e5213fc48218ec305b5434ef11e70ddaR6) cause an error when the request body is, for instance, an image?
> 
> I am not sure that it would cause an error (did a quick search and [the docs](https://developer.mozilla.org/en-US/docs/Web/API/Request/text) do not appear to mention that), maybe it will simply be a utf-8 string representation of the bytes. I do agree that that may not be the best way to handle those cases, feel free to open an issue about that.

From my testing, you're right, it doesn't error. However, it still means we won't be able to upload images (or other binary formats) properly to our backend. I'll open the issue.

BrunoRosendo:
I can't test because of this error:
```
Error: Invalid export 'HEAD' in src/routes/api/[...endpoint]/+server.ts (valid exports are GET, POST, PATCH, PUT, DELETE, OPTIONS, prerender, trailingSlash, config, or anything with a '_' prefix)
```

Did you encounter it?

limwa:
> I can't test because of this error:
> ```
> Error: Invalid export 'HEAD' in src/routes/api/[...endpoint]/+server.ts (valid exports are GET, POST, PATCH, PUT, DELETE, OPTIONS, prerender, trailingSlash, config, or anything with a '_' prefix)
> ```
> 
> Did you encounter it?

That issue has been fixed in SvelteKit v1.22.0 (https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md).

I just noticed that we are using the dependency "@sveltejs/kit" with the tag "next". As per [this page](https://www.npmjs.com/package/@sveltejs/kit?activeTab=versions), that will resolve to version "1.0.0-next.589". We might want to change the tag in the package.json file to "latest" or something else as that version is very outdated (7 months old).

BrunoRosendo (approved):
lgtm, the issue should be solved after bumping sveltekit version

PR #220, opened 2023-08-09
Timeline Component

### LINKED ISSUE #19: members: team page
Use the work developed in #1, #89  and the mockups (examples below) to implement the team page. Note that the sections come from the members' roles and are **not** static (consult docs or speak with a backend person to understand how to use the API).

When the user hovers over a member, the **grey** overlay should appear and show extra information about the person.

The issue should be composed of the following tasks:
- [ ] Implement PC view
- [ ] Implement mobile view

Image
Image

### LINKED ISSUE #282: members, events: timeline component
Following the completion of the priority version of the team #19  and events page #14 for the MVP, a common timeline component should be developed, according to the mocks below.

The addition of the possibility of descriptions along the timeline should be discussed, in order to try to generalize this component to the project's detail page #18 .

Image

Image

Image

Image

LuisDuarte1:
Closes #282 

# Screenshots

# Review checklist

- [ ] Behavior is as expected


LuisDuarte1:
The HorizontalTimeline for now: 
image

I've changed the design from the mockups (because it's more intuitive having the arrows for interacting with the timeline imo). wdyt? @MRita443 

Also, should I change this PR to be the HorizontalTimeline only? because I think that #163 depends on it

LuisDuarte1:
There's some things that I still have to consider:

- Should we change the `displayedSize` according to width? (must be done on js though)
- Should we have transition animations? (not sure what kind of animations would be suitable for this)
- What color should we use to indicate that you can't go futher right/left? 
- Testing the component :sweat_smile:

MRita443:
> 
> The HorizontalTimeline for now: 
> image
> 
> I've changed the design from the mockups (because it's more intuitive having the arrows for interacting with the timeline imo). wdyt? @MRita443 
> 
> Also, should I change this PR to be the HorizontalTimeline only? because I think that #163 depends on it
> 
> 
> 

I also agree this way is more intuitive for knowing there are more years to scroll through, but let's ask @NIAEFEUP/design-website for an opinion too!

If you see the PR is getting large or that it might take you a bit to finish the whole page go ahead and separate it, but I don't think #163 will be worked on for now.

Keep up the good work :)

PR #295, opened 2024-11-16
initial user info page

### LINKED ISSUE #284: user: info page
Implement the priority version of the user's personal page, according to mocks below.

Image

peucastro:
Closes #284

### **Changes Implemented:**  
- **User Info Page**: Implemented priority version with backend integration.  
- **New Components**: Added `IsActiveBadge` and `SocialMediaIcon` local components.  
- **Edit Button**: Created a global `EditButton` component.  
- **Error Page Enhancements**:  
  - Updated layout for better user experience.  
  - Added Storybook entry for the error page.  
- **Type Update**: Renamed `gitHub` to `github` in `TeamMember` type to match with backend.  

## Screenshots
Desktop version             |  Mobile version
:-------------------------:|:-------------------------:
  |  

# Review checklist

- [ ] Behavior is as expected


src/routes/(app)/user/[id]/+page.svelte
```diff
@@ -0,0 +1,139 @@
...
+  import Icons from '@/lib/components/icons/icons';
+
+  import type { TeamMember } from '@/types/team-member';
+  import Snackbar from '@/routes/(app)/_components/layout/notifications/snackbar.svelte';
+  import { page } from '$app/stores';
+  import { goto } from '$app/navigation';
+
+  export let teamMember = getMember() as TeamMember;
+
+  async function getMember() {
```
Comment (rubuy-74) [inline src/routes/(app)/user/[id]/+page.svelte]: Instead of directly making calling fetch in the page, you should use a load function in another file to load the data, as it is described in the documentation (https://svelte.dev/docs/kit/load).

src/routes/(app)/user/[id]/+page.svelte
```diff
@@ -0,0 +1,139 @@
...
+            </div>
+            {#if teamMember.bio}
+              <p
+                class="text-md lg:text-md mb-6 text-center sm:text-left sm:text-base xl:text-lg 2xl:text-xl"
+              >
+                {teamMember.bio}
+              </p>
+            {/if}
+            <div class="flex justify-center gap-3 sm:justify-start">
+              {#if teamMember.github}
```
Comment (rubuy-74) [inline src/routes/(app)/user/[id]/+page.svelte:64]: As we discussed, since this is used in a different place, it would be good to have it as a separated component because there could be more uses of this in the future.

src/routes/(app)/user/[id]/+page.svelte
```diff
@@ -0,0 +1,139 @@
...
+                    {:else}
+                      <Icon src={Icons.Globe} color="white" size="100%" />
+                    {/if}
+                  </a>
+                {/each}
+              {/if}
+            </div>
+          </div>
+        </div>
+        {#if showErrorMessage}
```
Comment (rubuy-74) [inline src/routes/(app)/user/[id]/+page.svelte]: This is something that could be handled globally on the website. Nothing wrong with this, but its something we can discuss in the next standup

rubuy-74 (changes requested):
Everything looks great, there are only small things that could be a little different. :smile:

src/routes/(app)/user/[id]/+page.svelte
Comment (peucastro) [reply src/routes/(app)/user/[id]/+page.svelte]: Thank you for providing the resources for these changes. I have applied the updates as suggested. Could you please review the changes and confirm if everything is correct?

src/routes/(app)/user/[id]/+page.svelte
Comment (peucastro) [reply src/routes/(app)/user/[id]/+page.svelte]: I totally agree. I have removed it for now, and we can discuss this further at the next meeting.

src/routes/(app)/user/[id]/+page.svelte
Comment (peucastro) [reply src/routes/(app)/user/[id]/+page.svelte]: Since we were not able to mock `PageData` at this page's Storybook, we've decided to keep the fetch function at the page file itself. Thus, I'll mark this as resolved.

src/routes/(app)/user/[id]/+page.svelte
```diff
@@ -0,0 +1,103 @@
...
+            <a href="https://www.lipsum.com/">
+              <Icon src={Icons.Edit} color="white" size="100%" />
+            </a>
+          </div>
+          <button
+            class="w-12 rounded-md bg-muted-red-500 p-3 lg:w-12 xl:w-12"
+            on:click={() => logout()}
+          >
+            <Icon src={Icons.Logout} color="white" size="100%" />
+          </button>
```
Comment (MRita443) [inline src/routes/(app)/user/[id]/+page.svelte]: Maybe you could use the `Button` component here for consistency? If you need you can add an optional `onClick` argument to it and pass a function. If you could also abstract `EditButton` into its own component it'd be great because it's used in a lot of pages!

src/routes/(app)/user/[id]/+page.ts
```diff
@@ -0,0 +1,11 @@
+import { error } from '@sveltejs/kit';
+import type { PageLoad } from './$types';
+
+export const load: PageLoad = async ({ fetch, params }) => {
+  const res = await fetch(`/api/accounts/${params.id}`);
+  if (res.status != 200) error(res.status, 'User not found');
```
Comment (MRita443) [inline src/routes/(app)/user/[id]/+page.ts]:
```suggestion
  if (!res.ok) error(res.status, 'User not found');
```

src/routes/(app)/user/_components/social-media-icon.stories.ts
MRita443, line 1: There should be at least one example here that uses `iconPath`

src/routes/(app)/+error.svelte
MRita443, line 1: (A little out of scope of this PR but) could you add this page to storybook too?

MRita443 (changes requested):
It's looking great! Nearly done, just small organizational and test things

src/routes/(app)/user/[id]/+page.svelte
Comment (peucastro) [reply src/routes/(app)/user/[id]/+page.svelte]: Wouldn't it be better if, instead of creating a new component, we modify the existing `Button` so it can receive an image as a parameter? Creating a new component just for the `EditButton` seems a little redundant, since most of the logic would be extracted from the `Button` component anyway. Alternatively, we could create a new button component that receives just images, not text. Which option do you think it's better?

src/routes/(app)/+error.svelte
peucastro: I was trying to add it now, but since Storybook doesn't support the `$app/stores` module natively, it can get a little bit tricky to implement this. Instead of loading the error page, it just displays an internal error (due to the bad http status). I've tried using MSW to bypass this but couldn't get it to work. We can discuss this better at tomorrow's meeting!

peucastro: Try to use sveltekit_experimental mock such as in #303 and described at [Storybook's docs](https://storybook.js.org/docs/get-started/frameworks/sveltekit#how-to-mock).

src/routes/(app)/user/[id]/+page.svelte
Comment (MRita443) [reply src/routes/(app)/user/[id]/+page.svelte]: As we talked in the meeting (but just to immortalize here), yes you are absolutely right! But, as the edit button is used so many times throughout the design, I think it's worth it to guarantee that every page has the same consistent button and no slight changes in color, size or icon

src/routes/(app)/user/[id]/+page.svelte
```diff
@@ -0,0 +1,106 @@
...
+                <div class="flex items-center rounded-2xl bg-green-200/20 px-5 py-2">
+                  <div class="mr-2 h-4 w-4 rounded-full bg-green-200"></div>
+                  <p class="text-sm font-semibold sm:text-base">Membro ativo</p>
+                </div>
+              {:else}
+                <div class="flex items-center rounded-2xl bg-red-200/20 px-5 py-2">
+                  <div class="mr-2 h-4 w-4 rounded-full bg-red-500"></div>
+                  <p class="text-sm font-semibold sm:text-base">Membro não ativo</p>
+                </div>
+              {/if}
```
Comment (MRita443) [inline src/routes/(app)/user/[id]/+page.svelte]: It'd be great if this could be a component too! I'd probably place it in `/user/_components` because it doesn't seem reusable in other pages at the moment but at least the logic is extracted and in storybook

Comment (MRita443) [reply src/routes/(app)/user/[id]/+page.svelte]: On that topic, I'd probably create a `src/lib/components/buttons` folder for organization too!

src/lib/components/edit-button.stories.ts
```diff
@@ -0,0 +1,31 @@
+import EditButton from './edit-button.svelte';
+
+export default {
+  title: 'Atoms/Buttons/EditButton',
```
MRita443:
```suggestion
  title: 'Atoms/Buttons/Edit Button',
```

MRita443 (approved):
Great job! This PR is great :) I left some small comments for if you still have patience but if not this is done and great as is already

src/routes/(app)/user/[id]/+page.svelte
Comment (peucastro) [reply src/routes/(app)/user/[id]/+page.svelte]: Thanks for the suggestions!

rubuy-74 (approved):
LGTM :rocket:

PR #303, opened 2024-11-30
highlighted the current page item on navebar

### LINKED ISSUE #300: navbar: highlight current page
We should change our navbar to highlight the page we are currently visiting accordingly.
For example, when visiting the team page, the navbar should highlight `Equipa`:

Image

We should aim to develop a general solution that doesn't involve changing the navbar styling in each page individually.

Isaac3105:
Closes #300

Now, if your current page is one of the pages from navebar, the corresponding item from the navebar will be highlighted.
image

# Review checklist

- [ ] Behavior is as expected


src/routes/(app)/_components/layout/navbar.svelte
```diff
@@ -1,3 +1,12 @@
+<script>
+  import { onMount } from 'svelte';
+  let currentUrl = '';
+
+  onMount(() => {
+    currentUrl = window.location.href;
+  });
+</script>
```
MRita443: This is a good solution but an even easier one would be to use the functionalities from SvelteKit directly! Here is how you can get what's after `http://localhost:3000/`:
```suggestion
<script>
  import { page } from '$app/stores';

  $: currentPage = $page.url.pathname;
</script>
```

src/routes/(app)/_components/layout/navbar.svelte
```diff
@@ -9,9 +18,17 @@
...
   </div>
   <div class="flex justify-end gap-7">
-    <a href="#/">Equipa</a>
-    <a href="#/">Projetos</a>
-    <a href="#/">Eventos</a>
-    <a href="#/">Contactos</a>
+    <a class:active={currentUrl === "http://localhost:3000/#/team/"} href="#/team/"><p class = "font-bold">Equipa</p></a>
+    <a class:active={currentUrl === "http://localhost:3000/#/projects/"} href="#/projects/"><p class = "font-bold">Projetos</p></a>
+    <a class:active={currentUrl === "http://localhost:3000/#/events/"} href="#/events/"><p class = "font-bold">Eventos</p></a>
+    <a class:active={currentUrl === "http://localhost:3000/#/contacts/"} href="#/contacts/"><p class = "font-bold">Contactos</p></a>
```
MRita443: The `#` notation is used to reference an id of a HTML element inside a page, so this is not the intended use case! The previous one was a placeholder :) If you actually wanted to use it, you could do `/contacts/#button` (fictional example), so that once you clicked it you'd be redirectioned to the contacts page and scrolled down automatically to the button. Here you should do `/contacts`, `/team` etc. though!

MRita443: I also think we can avoid having this much repetition and do a more general code, here is a hint as to how I think that could be done:  
`  const links = [
    { href: '/team', label: 'Equipa' }, ...
  ];`

src/routes/(app)/_components/layout/navbar.svelte
```diff
@@ -9,9 +18,17 @@
...
   </div>
 </nav>
+
+<style>
+  .active {
+    background-color: rgba(116, 20, 20, 0.4);
+    padding: 2px 3px;
```
MRita443: A little more padding on the left and right would look good :) Also for it to be more readable you can use `theme('colors.muted-red.400')` (example), as to use the colors we already have defined, if you don't know what they are check out `tailwind.config.cjs`

MRita443 (changes requested):
Good job! I left some suggestion as to how you can make this even better :)

src/routes/(app)/_components/layout/navbar.svelte
Isaac3105: Done!

src/routes/(app)/_components/layout/navbar.svelte
Isaac3105: Done!

src/routes/(app)/_components/layout/navbar.svelte
Isaac3105: DONEEEEE!!!!!

src/routes/(app)/_components/layout/navbar.svelte
```diff
@@ -1,3 +1,17 @@
+<script>
+  // @ts-expect-error Import is as expected but throws error
+
+  import { page } from '$app/stores';
+
+  $: currentPage = $page.url.pathname;
+  const links = [
```
rubuy-74, line 7: For the pages that are not ready, this is linking for a 404 page. Shouldn't it be better to redirect to "#", as it was before?

src/routes/(app)/_components/layout/navbar.svelte
```diff
@@ -9,9 +23,18 @@
...
-    <a href="/contacts">Contactos</a>
+    {#each links as { href, label, pageComp }}
+      <a class:active={currentPage === pageComp} {href}>
+        <p class="font-bold">{label}</p>
+      </a>
+    {/each}
   </div>
 </nav>
+
+<style>
```
rubuy-74: Since we are using Tailwind to define the style of html components, you should avoid using pure CSS.

src/routes/(app)/_components/layout/navbar.svelte
```diff
@@ -9,9 +23,18 @@
     </a>
   </div>
   <div class="flex justify-end gap-7">
-    <a href="/#/">Equipa</a>
-    <a href="/#/">Projetos</a>
-    <a href="/#/">Eventos</a>
-    <a href="/contacts">Contactos</a>
+    {#each links as { href, label, pageComp }}
+      <a class:active={currentPage === pageComp} {href}>
```
rubuy-74: Here you could use a svelte conditional or a ternary to render the different items on the navbar.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator
https://svelte.dev/tutorial/svelte/if-blocks

rubuy-74 (changes requested):
Great job, there are just some small things that could be a little different.

rubuy-74:
Another thing, now that the navbar has state, it should be nice to have some storybook tests. There are some examples already in the codebase but you can also use the documentation as reference.

https://storybook.js.org/docs/writing-tests

src/routes/(app)/_components/layout/navbar.svelte
Isaac3105: Fixed. Pages that are still not ready are now redirected to "#".

src/routes/(app)/_components/layout/navbar.svelte
Isaac3105: Fixed. Implemented the style in class.

src/routes/(app)/_components/layout/navbar.svelte
Isaac3105: Fixed. Implemented if blocks.

src/routes/(app)/_components/layout/navbar.svelte
```diff
@@ -1,5 +1,16 @@
 <script>
+  // @ts-expect-error Import is as expected but throws error
+
   import { onMount } from 'svelte';
+  import { page } from '$app/stores';
```
MRita443:
```suggestion
  import { onMount } from 'svelte';
    // @ts-expect-error Import is as expected but throws error
  import { page } from '$app/stores';
```
That comment needs to be above the line with the error for it to work :)

src/routes/(app)/_components/layout/navbar.svelte
```diff
@@ -31,11 +42,18 @@
...
   </div>
-  <div class="flex justify-end gap-7">
-    <a href="/#/">Equipa</a>
-    <a href="/#/">Projetos</a>
-    <a href="/#/">Eventos</a>
-    <a href="/contacts">Contactos</a>
+  <div class="flex items-center justify-end gap-7">
+    {#each links as { href, label, pageComp }}
+      {#if currentPage === pageComp}
+        <a {href} class="rounded bg-muted-red-400 p-2" data-testid={label}>
```
MRita443: For consistency maybe the testid should be the label but lowercased

src/routes/(app)/_components/layout/navbar.svelte
```diff
@@ -1,5 +1,16 @@
 <script>
+  // @ts-expect-error Import is as expected but throws error
+
   import { onMount } from 'svelte';
+  import { page } from '$app/stores';
+
+  $: currentPage = $page.url.pathname ?? '/';
+  const links = [
+    { href: '#', label: 'Equipa', pageComp: '/team' },
```
MRita443, line 8: What does `pageComp` stand for?

MRita443 (changes requested):
Great job, nearly done!

src/routes/(app)/_components/layout/navbar.svelte
Isaac3105: done!

src/routes/(app)/_components/layout/navbar.svelte
Isaac3105: done!

src/routes/(app)/_components/layout/navbar.svelte
Isaac3105: I was thinking something like: I need to **comp**are the current page with each **page** available in the navbar... I agree, it's kind of a stupid name. 
If you want, I can change for something better.

MRita443 (approved):
Great job :)

rubuy-74 (changes requested):
Everything looks awesome. Just rebase the branch with the develop branch

rubuy-74 (approved):
LGTM :rocket:

PR #307, opened 2024-12-08
feature: Edit Project page

### LINKED ISSUE #286: projects: info form page
Develop the priority version of a project's edition forms, according to mocks below.

Image

Image

DaltonDjoesman:
Closes #286

created a grid for the info form page


src/routes/(app)/projects/[id]/edit/+page.svelte
```diff
@@ -0,0 +1,154 @@
...
+                    Guardar Alterações
+                </p>
+            </div>
+            <div class="h-10 w-[10vw] bg-gray-500 rounded-xl flex items-center justify-center">
+                <p class="font-bold">
+                  Cancelar  
+                </p>
+            </div>
+        </div>
+    </div>
```
Comment (MRita443) [inline src/routes/(app)/projects/[id]/edit/+page.svelte]: For the buttons you can use the already existing `Button` component!

MRita443 (changes requested):
The page is looking great so far! :) It would be great if you could extend the bar below the "Editar Projeto" a little more like this:

image

I think when you integrate the forms components you will be able to simplify the code a bit. You should also add this page to Storybook, if you need any help just tell me :)

src/routes/(app)/projects/[id]/edit/+page.server.ts
```diff
@@ -0,0 +1,40 @@
...
+    const formData: FormData = await request.formData();
+    console.log(formData);
+    const id = params.id;
+    const googlePlay = formData.get('googleplay');
+    const appStore = formData.get('appstore');
+
+    console.log(googlePlay);
+    console.log(appStore);
+    console.log([googlePlay, appStore]);
+    console.log('---------------------------------------------------------------------');
```
Comment (rubuy-74) [inline src/routes/(app)/projects/[id]/edit/+page.server.ts]: We should avoid keeping "console.log" statements to get into production code. It is a very good tool for debugging, but please remove this and the remaining "console.log" statements.

src/routes/(app)/projects/[id]/edit/+page.server.ts
```diff
@@ -0,0 +1,40 @@
...
+      links: [{ url: googlePlay }, { url: appStore }],
+      addFile: formData.get('addFile')
+    };
+
+    const json = JSON.stringify(value);
+    const blob = new Blob([json], { type: 'application/json' });
+    const form = new FormData();
+    form.append('project', blob);
+
+    console.log(value);
```
Comment (rubuy-74) [inline src/routes/(app)/projects/[id]/edit/+page.server.ts]: Same as above.

src/routes/(app)/projects/[id]/edit/+page.svelte
```diff
@@ -0,0 +1,130 @@
...
+          label="Slogan"
+          name="slogan"
+          horizontal
+          isTextArea
+          placeholder="Insira o texto"
+          className="flex-col md:flex-row"
+          textGap="10"
+        />
+
+        <!--<div class="flex w-[50vw] flex-col gap-5 md:flex-row">
```
Comment (rubuy-74) [inline src/routes/(app)/projects/[id]/edit/+page.svelte]: Dead code should also be avoided. As this is not going to be executed it is best to remove it.

src/routes/(app)/projects/[id]/edit/+page.svelte
```diff
@@ -0,0 +1,130 @@
...
+          name="public"
+          horizontal
+          required
+          placeholder="Insira o texto"
+          className="flex-col md:flex-row"
+          textGap="10"
+        />
+
+        <div class="flex w-[50vw] gap-10">
+          <RadioButtons options={['Guardar Alteracoes', 'Cancelar']} horizontal />
```
Comment (rubuy-74) [inline src/routes/(app)/projects/[id]/edit/+page.svelte]: These butttons should have some functionality, namely editing the project or going back to the projects page.

src/routes/(app)/projects/[id]/edit/+page.svelte
```diff
@@ -0,0 +1,130 @@
...
+        </div>
+
+        <Button type="submit" color="primary" hoverColor="secondary" text="morra" />
```
Comment (rubuy-74) [inline src/routes/(app)/projects/[id]/edit/+page.svelte]: This button was only for testing purposes.

src/routes/(app)/projects/edit/[id]/+page.svelte
```diff
@@ -0,0 +1,130 @@
+<script>
```
Comment (rubuy-74) [inline src/routes/(app)/projects/edit/[id]/+page.svelte]: Another thing, users should be able to not edit some of the information about the project (i.e. keep it as it is). Therefore, no entry should be required, as the fields that were not updated should remain the same.

rubuy-74 (changes requested):
Great work, there are some few things that need changing.

Also, it is important to test if the update is actually doing something in the backend. While testing, there is no update on the projects page.

src/routes/(app)/projects/[id]/edit/+page.svelte
Comment (DaltonDjoesman) [reply src/routes/(app)/projects/[id]/edit/+page.svelte]: To address this feedback, could we replace each radio input with our Button component so that we can set its type (e.g. "submit", "reset" )

public/mockServiceWorker.js
MRita443, line 1: This wasn't changed on purpose right?

MRita443, line 1: This wasn't changed on purpose right?

src/lib/components/forms/file-input.stories.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export default {
+  title: 'Atoms/Forms/File Input',
+  component: FileInput,
+  parameters: {
+    layout: 'centered'
+  }
+};
+
+export const TestFileInput = {};
```
MRita443:
```suggestion
export default {
  title: 'Atoms/Forms',
  component: FileInput,
  parameters: {
    layout: 'centered'
  }
};

export const File_Input = {};
```
When you only have one example you don't need the extra "folder"

src/lib/components/forms/file-input.svelte
```diff
@@ -0,0 +1,35 @@
+<script>
+  export let name = '';
+
+  let files = [];
+  function append_file(event) {
+    files = Array.from(event.target.files);
+  }
+  function pop_file(fileRemove) {
```
MRita443:
```suggestion
  function appendFile(event) {
    files = Array.from(event.target.files);
  }
  function popFile(fileRemove) {
```

For consistency with other files

src/routes/(app)/(home)/+page.svelte
MRita443, line 1: This should be removed when the component is done

MRita443 (commented):
This review is more relative to the `FileInput` component (sorry in an attempt to simplify the PRs I accidentally joined them both here), the main thing besides the comments is the difference in looks, it should look more like this (but with the X), and maybe with a darker red on hover:

image

The no files text is also missing

src/lib/components/buttons/button.svelte
peucastro, line 1: Mind the colors used in the mockups (#286), you can reuse the colors that already exist

src/routes/(app)/projects/[id]/edit/+page.svelte
```diff
@@ -0,0 +1,90 @@
+<script>
+  import FileInput from '../../../../../lib/components/forms/file-input.svelte';
+  import LabelInput from '../../../../../lib/components/forms/label-input.svelte';
+  import PictureInput from '../../../../../lib/components/forms/picture-input.svelte';
+  import Button from '../../../../../lib/components/buttons/button.svelte';
```
Comment (peucastro) [inline src/routes/(app)/projects/[id]/edit/+page.svelte]: Always try to use absolute paths (e.g. `'@/lib/components/forms/label-input.svelte'`)

src/routes/(app)/projects/[id]/edit/+page.svelte
```diff
@@ -0,0 +1,90 @@
...
+</script>
+
+<div class="flex w-full flex-col items-center justify-around">
+  <div class="flex w-full justify-center border-b-4 border-red-500 p-2">
+    <h1 class="text-4xl font-bold">Editar projeto</h1>
+  </div>
+
+  <form method="POST" enctype="multipart/form-data" class="flex w-full flex-col items-center p-10">
+    <p class="mb-10 text-center text-2xl font-bold">Dados Gerais</p>
```
Comment (peucastro) [inline src/routes/(app)/projects/[id]/edit/+page.svelte]: Nitpick: here you should use a header like `h1` or `h2`

src/routes/(app)/projects/edit/[id]/+page.svelte
Comment (peucastro) [inline src/routes/(app)/projects/edit/[id]/+page.svelte:1]: In the form main body, most of the code is repeated across the inputs (same direction, class and text gap). To make the code more modularized, you can create a `_components` directory and export this input format as a component at the page

src/routes/(app)/projects/[id]/edit/+page.svelte
```diff
@@ -0,0 +1,90 @@
...
+          horizontal
+          placeholder="Insira o texto"
+          className="flex-col md:flex-row"
+          textGap="10"
+        />
+
+        <div class="flex flex-col gap-5 md:flex-row">
+          <p class="w-1/6 font-bold">Fotos</p>
+          <FileInput name="addFile" />
+        </div>
```
Comment (peucastro) [inline src/routes/(app)/projects/[id]/edit/+page.svelte]: This input is a bit out of alignment:

Image

src/routes/(app)/projects/[id]/edit/+page.svelte
```diff
@@ -0,0 +1,90 @@
...
+        />
+
+        <div class="flex w-[50vw] gap-10">
+          <Button type="submit" color="reddish" hoverColor="reddish" text="Guardar Alterações" />
+          <Button type="reset" color="reddish" hoverColor="reddish" text="Cancelar" />
+        </div>
+      </div>
+
+      <div class="order-1 flex flex-col items-center gap-5 self-center md:order-2 md:self-start">
+        <PictureInput name="image" text="logo" />
```
Comment (peucastro) [inline src/routes/(app)/projects/[id]/edit/+page.svelte]:
```suggestion
        <PictureInput name="image" text="Adicionar logo" />
```

src/routes/(app)/projects/[id]/edit/+page.svelte
```diff
@@ -0,0 +1,90 @@
...
+        <PictureInput name="imageMenu" text="imagem do menu" />
```
Comment (peucastro) [inline src/routes/(app)/projects/[id]/edit/+page.svelte]: What exactly is the menu image?

package-lock.json
```diff
@@ -2767,7 +2767,8 @@
         "mrmime": "^2.0.0",
         "sade": "^1.8.1",
         "set-cookie-parser": "^2.6.0",
-        "sirv": "^3.0.0"
+        "sirv": "^3.0.0",
+        "vitefu": "^1.0.6"
```
peucastro: Is this package being used?

peucastro (changes requested):
Overall, the page looks pretty solid! Just review the mockups at figma and adjust some small details. Some of them I have already pointed out. Also resolve some of the comments that Rita made, they still need to be addressed 😃

peucastro:
Another detail that I forgot to mention, this page should display a pop-up to confirm that the user really wants to discard the changes. The mockups for this are already at figma

DaltonDjoesman:
While working on this issue, the lint test failed due to an unused property horizontal in the forms/label-input.svelte component.

Although I didn't originally change this component, I have commented out the unused property to satisfy the linting rules.

src/routes/(app)/projects/edit/[id]/+page.svelte
```diff
@@ -0,0 +1,52 @@
...
+      <div class="order-2 flex w-full flex-col gap-5 md:order-1 md:w-1/2">
+        <ProjectFormsInput label="Title" name="title" value={project.title} />
+
+        <ProjectFormsInput label="Slug" name="slug" value={project.slug} />
+
+        <ProjectFormsInput label="GitHub" name="github" value={project.github} />
+
+        <ProjectFormsInput label="Slogan" name="slogan" isTextArea value={project.slogan} />
+
+        <ProjectFormsInput label="Descrição" name="description" value={project.description} />
```
Comment (peucastro) [inline src/routes/(app)/projects/edit/[id]/+page.svelte]: I think it would be better if the slogan input was a normal text input and the Description a Text Area

src/lib/components/buttons/button.svelte
```diff
@@ -13,7 +13,8 @@
     secondary: { main: 'bg-muted-red-500', hover: 'hover:bg-muted-red-500' },
     red: { main: 'bg-red-500', hover: 'hover:bg-red-500' },
     green: { main: 'bg-green-500', hover: 'hover:bg-green-500' },
-    blue: { main: 'bg-blue-500', hover: 'hover:bg-blue-500' }
+    blue: { main: 'bg-blue-500', hover: 'hover:bg-blue-500' },
+    reddish: { main: 'bg-taupe-200', hover: 'hover:bg-red-500' }
```
peucastro: This property is not being used

src/lib/components/buttons/button.svelte
```diff
@@ -26,7 +27,9 @@
   {type}
   class="flex {colorList[color]['main']} {colorList[hoverColor][
     'hover'
-  ]} justify-center font-raleway font-bold text-white {widthList[width]} min-w-0 rounded px-4 py-2"
+  ]} items-center justify-center font-raleway font-bold text-white {widthList[
+    width
+  ]} min-w-0 rounded px-4 py-2"
```
peucastro: Is this change necessary?

src/lib/components/forms/file-input.stories.ts
peucastro, line 1: Make sure to add some tests here

src/lib/components/forms/file-input.svelte
```diff
@@ -0,0 +1,87 @@
+<script lang="ts">
+  interface Props {
+    name?: string;
+    value?: (string | File)[];
+  }
+
+  let { name = '', value = $bindable([]) }: Props = $props();
+
+  let files = $state(value);
```
peucastro: This creates a non-reactive local copy of the prop. In Svelte, initializing `$state` with a prop only runs once. Ff the parent's value changes later, this local files state will not update. To maintain a reactive link with the parent, you should remove this local state and operate on value directly

src/lib/components/forms/file-input.svelte
```diff
@@ -0,0 +1,87 @@
...
+    if (target.files) {
+      const newFiles = Array.from(target.files);
+      files = [...files, ...newFiles];
+      updateInputElement();
+    }
+  }
+
+  function removeFile(fileToRemove: string | File) {
+    if (typeof fileToRemove === 'string') {
+      filesDeleted.push(fileToRemove);
```
peucastro:
```suggestion
      filesDeleted = [...filesDeleted, fileToRemove];
```

src/lib/components/forms/file-input.svelte
```diff
@@ -0,0 +1,87 @@
...
+    bind:this={inputElement}
+  />
+
+  {#each filesDeleted as deletedUrl (deletedUrl)}
+    <input type="hidden" name="{name}_to_delete" value={deletedUrl} />
+  {/each}
+
+  {#if files.length}
+    <div class="flex flex-wrap gap-2">
+      {#each files as file (getFileName(file))}
```
peucastro: Avoid using the filename as a key, since duplicate names can break the DOM. It is preferable to use `(file)` as the key to track the object reference instead

src/lib/components/forms/label-input.svelte
```diff
@@ -22,26 +22,26 @@
     type = 'text',
     placeholder = '',
     isTextArea = false,
-    horizontal = false,
+    //horizontal = false,
```
peucastro:
```suggestion
    horizontal = false,
```

src/lib/components/forms/label-input.svelte
```diff
@@ -22,26 +22,26 @@
...
     required = false,
     minlength = undefined,
     maxlength = undefined,
     ...rest
   }: Props = $props();
 </script>
 
-<div class="flex flex-{horizontal ? 'row' : 'col'}">
+<div class="flex flex-col md:flex-row">
```
peucastro:
```suggestion
<div class="flex flex-{horizontal ? 'row' : 'col'}">
```

src/lib/components/forms/picture-input.svelte
```diff
@@ -7,12 +7,12 @@
   interface Props {
     text: string;
     name?: string;
+    value?: string;
   }
 
-  let { text, name = 'profilePicture' }: Props = $props();
-  let image: string | undefined = $state();
+  let { text, name = 'profilePicture', value = $bindable() }: Props = $props();
+  let image: string | undefined = $state(value);
```
peucastro: Just like in the file input, this creates a local copy of value that isn't reactively linked to the parent. Since value is `$bindable()`, you should operate on value directly

src/routes/(app)/projects/edit/[id]/+page.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
+  import ProjectFormsInput from '../../_components/project-forms-input.svelte';
+  import type { PageData } from './$types';
+  import FileInput from '@/lib/components/forms/file-input.svelte';
+  import PictureInput from '@/lib/components/forms/picture-input.svelte';
+  import Button from '@/lib/components/buttons/button.svelte';
+  import type { Project } from '@/types/project';
+
+  let { data }: { data: PageData } = $props();
+  const project: Project = data.project;
```
Comment (peucastro) [inline src/routes/(app)/projects/edit/[id]/+page.svelte]: With the suggested changes, if you want the form to be reactive, you should wrap the prop in a `$state` rune. Since it is currently a `const`, any changes made inside the components won't be tracked by this page.

src/routes/(app)/projects/edit/[id]/+page.svelte
```diff
@@ -0,0 +1,52 @@
...
+</script>
+
+<div class="flex w-full flex-col items-center justify-around">
+  <div class="flex w-full justify-center border-b-4 border-red-500 p-2">
+    <h1 class="text-4xl font-bold">Editar projeto</h1>
+  </div>
+
+  <form method="POST" enctype="multipart/form-data" class="flex w-full flex-col items-center p-10">
```
Comment (peucastro) [inline src/routes/(app)/projects/edit/[id]/+page.svelte]: The form doesn't handle loading states or errors from the server. Consider adding `use:enhance`. This will prevent a full page reload and allow you to use form to show success or error messages to the user. There is an example on how to implement this in `routes/(app)/admin/_components/technologies.svelte`

src/routes/(app)/projects/edit/[id]/+page.ts
```diff
@@ -0,0 +1,10 @@
+import { error } from '@sveltejs/kit';
+import type { PageLoad } from './$types';
+
+export const load: PageLoad = async ({ fetch, params }) => {
+  let res = await fetch(`/api/projects/${params.id}`);
+  if (!res.ok) error(res.status, 'Project not found');
+  const project = await res.json();
+  //console.log(project)
```
Comment (peucastro) [inline src/routes/(app)/projects/edit/[id]/+page.ts]:
```suggestion
```

peucastro (changes requested):

src/lib/components/forms/label-input.svelte
DaltonDjoesman: @peucastro , I don´t understand why this method is better than the other one, how will the page know if it is in a mobile or in a pc without using md?
with the substitution the mobile version appears unformatted

peucastro: Removing this will make the horizontal parameter unused, which breaks the layout for components that rely on it (like `ProjectFormsInput`). Since this is a shared property, we need to keep the parameter and its logic intact. Could we refactor your change to work alongside the existing layout toggle?

src/routes/(app)/projects/edit/[id]/+page.svelte
```diff
@@ -0,0 +1,73 @@
...
+  let { data }: { data: PageData } = $props();
+  let project: Project = data.project;
+</script>
+
+<div class="flex w-full flex-col items-center justify-around">
+  <div class="flex w-full justify-center border-b-4 border-red-500 p-2">
+    <h1 class="text-4xl font-bold">Editar projeto</h1>
+  </div>
+
+  <form
```
Comment (peucastro) [inline src/routes/(app)/projects/edit/[id]/+page.svelte:65]: There is currently no client-side validation of whether the fields in the forms are valid or not. This will result in a silent error in the backend-side.

For example, if i try to submit the project with an empty description (this operation is not allowed by the backend), the API call will fail but the UI shows no sign of failure.

src/routes/(app)/projects/edit/[id]/+page.ts
```diff
@@ -0,0 +1,9 @@
+import { error } from '@sveltejs/kit';
+import type { PageLoad } from './$types';
+
+export const load: PageLoad = async ({ fetch, params }) => {
+  let res = await fetch(`/api/projects/${params.id}`);
```
Comment (peucastro) [inline src/routes/(app)/projects/edit/[id]/+page.ts]: The edit page currently expects a numeric project ID in the URL (`/projects/edit/1`), but it can also be accessed using the slug (`/projects/edit/nijobs`, for example). The issue is that when using the slug, the backend will reject the requests made using it, since it is expecting an id.

```
Failed to convert value of type 'java.lang.String' to required type 'long'; For input string: "nijobs"
```

Investigate how to deal with this and which URL would be more appropriate (/project/edit/nitsig, /project/nitsig/edit, /project/edit/1 ...)

RIP NIJobs 🕊️

src/lib/components/forms/file-input.svelte
```diff
@@ -0,0 +1,85 @@
...
+      {#each value as file (file)}
+        <div class="file-item flex min-w-0 items-center gap-2 rounded bg-taupe-200 px-2 py-2">
+          <span class="text-sm font-bold text-rose-950">{getFileName(file)}</span>
+          <button
+            type="button"
+            onclick={() => removeFile(file)}
+            class="rounded bg-red-200 px-2 font-bold text-black hover:bg-red-500"
+          >
+            X
+          </button>
```
peucastro, line 126: Here it would be better to use an Icon and include an aria-label for better accessibility

src/lib/components/forms/file-input.svelte
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  function appendFile(event: Event) {
+    const target = event.target as HTMLInputElement;
+    if (target.files) {
+      const newFiles = Array.from(target.files);
+      value = [...value, ...newFiles];
+      updateInputElement();
+    }
+  }
```
peucastro, line 79: There’s no check for accepted file types, file size, or limiting the number of files. Please add a client-side validation so users can't upload unsupported or overly large files.

src/lib/components/forms/file-input.svelte
```diff
@@ -0,0 +1,85 @@
...
+      }
+    });
+    inputElement.files = dataTransfer.files;
+  }
+
+  function appendFile(event: Event) {
+    const target = event.target as HTMLInputElement;
+    if (target.files) {
+      const newFiles = Array.from(target.files);
+      value = [...value, ...newFiles];
```
peucastro: Will this allow files with the same name? If so, you need to add some logic to prevent the same file from being added multiple times.

peucastro (changes requested):

PR #308, opened 2024-12-11
added navbar background that hides content under

### LINKED ISSUE #291: layout: fix content scrolling behind navbar
Currently, the `navbar` is set to `fixed`, so it doesn't occupy space in the layout. As a result, when content overflows vertically, it appears behind the `navbar`, and the page content is not properly centered. While removing the fixed positioning would fix this issue, we need the navbar to stay visible while scrolling. We should research solutions for this.

_Example on login page 680x410:_
Image

_Desired outcome:_
Image

RuiSoares333:
Closes #291 

Now the content gets hidden behind the Navbar.

# Screenshots
| Before | After |
| ------- | ------ |
|383137385-c183ccdc-1f40-45c1-8e05-b6810871489b | image |

# Review checklist

- [ ] Behavior is as expected

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -22,6 +22,14 @@ body {
     linear-gradient(180deg, rgba(56, 0, 2, 1), rgba(97, 14, 17, 1));
 }
 
+.bg-ni-navbar {
+  background-image: linear-gradient(
+    180deg,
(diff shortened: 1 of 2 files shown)


MRita443 (approved):
Great job!

PR #309, opened 2025-01-06
Fix/bg color login form

### LINKED ISSUE #292: layout: fix background color on login form fields
The form fields color in the login page are wrong for me at least.   

image

PedroOliveira42:
Closes #292 

Altered the login forms to not use the browsers autofill style changes.

## Before
image

## After
image

# Potencial Problems
This will make it so that every input field with autofill will have a pink background and letters, will most likely not become an issue though because few fields with autofill options will have diferent background options. However cresting a custom class for pink autofill inputs could also be done.

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -33,9 +33,14 @@
     overflow: auto;
   }
 
-  :global(input:-webkit-autofill,input:-webkit-autofill, input:-webkit-autofill:hover, input:-webkit-autofill:focus, input:-webkit-autofill:active) {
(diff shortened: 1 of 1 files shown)


src/routes/(app)/+layout.svelte
```diff
@@ -32,4 +32,15 @@
     flex-direction: column;
     overflow: auto;
   }
+
+  :global(
```
rubuy-74, line 40: Just a minor thing, it would be good to have a little bit more context. For people that could end up having the problems that you mentioned in the PR's description.

rubuy-74 (changes requested):

rubuy-74 (approved):
Nice work :rocket:

PR #310, opened 2025-01-20
Feature/storybook mock api

### LINKED ISSUE #305: tests: mock URL params and API calls
For the frontend's storybook, we should find a systematic way to mock URL params and API calls. We can attempt to use Vite or other Storybook resources for this. We should research the options and discuss them.

PedroOliveira42:
Closes #305 

Added MSW to Storybook making it possible to mock API calls.

## Storybook result

image

#### Notes:
 To make possible future Storybook tests instead of using 'window.location' as a redirect function use 'goto'.

Even after using the linter, '/public/mockServiceWorker.js' would cause conflicts with CI, since it is a file created by MSW and not likely to be edited by us added it to '.prettierignore'


MRita443 (approved):
Great job!

PR #379, opened 2025-12-07
Feature/loading state

### LINKED ISSUE #302: loading: consistent loading page/animation
We should integrate a loading animation on all of our pages when fetching data. We could do this individually, but it would be interesting to create a fetch wrapper and use a store to do that automatically in our layout for all pages.

PedroLunet:
Closes #302 

### Description of the changes proposed in the pull request

This PR introduces a Global Loading Animation system to improve UX during data fetching and page transitions.

**1. The "Hybrid" Implementation Strategy:**
Instead of wrapping individual `load` functions, I implemented a centralized solution in `src/routes/+layout.svelte` that reacts to two states:
* **Automatic:** Listens to SvelteKit's built-in `$navigating` store for page transitions.
* **Manual:** Introduced a new `isGlobalLoading` store (and a `loadingWrapper` helper) to trigger the loader during client-side actions (e.g., delete buttons, form submissions) without navigating.

**2. UX Improvements (Debouncing):**
To prevent the loader from "flashing" briefly on fast connections or during SvelteKit's hover-prefetching, I implemented a **delay threshold** (e.g., 150ms). The loader only appears if the data fetching takes longer than this threshold.

**3. Visuals:**
Created a new `GlobalLoader` component using a custom CSS animation that separates rotation from scaling. This creates a "breathing/pulsing" effect (`scale` 2s) combined with a snappy rotation (`spin` 2s), creating a high-end feel while handling `z-index` safety (z-9999) and accessibility (`prefers-reduced-motion`).

### Technical Note on SSR vs. Client-Side Behavior
**Current Behavior:** The loader appears on client-side navigation (clicking links) but **does not** appear on a hard refresh (F5) or direct URL access.

**Reasoning:** We are currently keeping `export const ssr = true`. On a hard refresh, the server fetches the data and delivers the HTML fully rendered. This provides better SEO and faster "First Contentful Paint" than showing a spinner.
* *Alternative considered:* We could force the loader on reload by setting `ssr = false`, but this was decided against to preserve performance and SEO benefits.

# Screenshots

https://github.com/user-attachments/assets/dfec312e-6602-488d-b6ad-74fb49abb60c

https://github.com/user-attachments/assets/75ffa65c-deef-4696-8513-8d1a764f8e55

# Review checklist

- [ ] Behavior is as expected


peucastro (approved):
Looks amazing! ✅

PR #382, opened 2025-12-17
Team Page

### LINKED ISSUE #19: members: team page

franpts2:
Closes #19

Implements the new Team Page route. This update introduces dynamic team data fetching from the backend and a fully responsive layout tailored for both desktop and mobile users.

**Changes:**
- New Route & Data: Added the /team page, which fetches current team sections from the /api/generations/latest endpoint.
- Responsive UI/UX:
	- Desktop: Role-based sections rendered in a clean grid layout.
	- Mobile: Integrated a swipeable/clickable section carousel for better navigation on small screens.

# Screenshots
## Desktop
image

Screenshot From 2026-03-01 15-42-02

## Mobile
image

# Review checklist

- [ ] Behavior is as expected


src/routes/(app)/team/+page.ts
```diff
@@ -0,0 +1,23 @@
...
+import type { PageLoad } from './$types';
+import type { TeamMember } from '@/types/team-member';
+
+type Section = {
+  section: string;
+  accounts: TeamMember[];
+};
+
+export const load: PageLoad = async ({ fetch }) => {
+  const res = await fetch(`/api/generations/23-24`);
```
peucastro:
```suggestion
  const res = await fetch(`/api/generations/latest`);
```

src/routes/(app)/team/+page.svelte
```diff
@@ -0,0 +1,303 @@
...
+  import HexagonGrid from '$lib/components/hexagons/hexagon-grid.svelte';
+
+  interface Account {
+    [key: string]: unknown;
+  }
+
+  interface Section {
+    section: string;
+    accounts: Account[];
+  }
```
peucastro: Here it would be better to use the `TeamMember` type directly

src/routes/(app)/team/+page.svelte
```diff
@@ -0,0 +1,303 @@
...
+        ? `-webkit-mask-image: linear-gradient(to right, transparent 0%, transparent calc(50% - ${maskLeftPx}px), black calc(50% - ${maskLeftPx}px), black calc(50% + ${maskRightPx}px), transparent calc(50% + ${maskRightPx}px), transparent 100%); mask-image: linear-gradient(to right, transparent 0%, transparent calc(50% - ${maskLeftPx}px), black calc(50% - ${maskLeftPx}px), black calc(50% + ${maskRightPx}px), transparent calc(50% + ${maskRightPx}px), transparent 100%);`
+        : `-webkit-mask-image: linear-gradient(to right, transparent 0%, transparent 15%, black 15%, black 85%, transparent 85%, transparent 100%); mask-image: linear-gradient(to right, transparent 0%, transparent 15%, black 15%, black 85%, transparent 85%, transparent 100%);`}
+    >
+      <div
+        bind:this={containerRef}
+        class="absolute left-1/2 flex h-full items-center transition-transform ease-in-out"
+        style="transform: translateX({slidePosition}px); transition-duration: {transitionDuration}ms"
+      >
+        {#each visibleSections as section, i (section.isDuplicate ? section.name + '_dup_' + i : section.name)}
+          {#if section.accounts.length > 0}
```
peucastro: Here we can add a default message if no team members are found

src/routes/(app)/team/+page.svelte
```diff
@@ -0,0 +1,303 @@
...
+          <HexagonGrid
+            items={section.accounts}
+            cols={4}
+            orientation="horizontal"
+            component={TeamMemberHexagon}
+          />
+        </div>
+      </div>
+    {/if}
+  {/each}
```
peucastro: Consider extracting this into a separate component to avoid code duplication

peucastro (changes requested):

src/routes/(app)/team/+page.svelte
```diff
@@ -0,0 +1,314 @@
...
+    if (openSection === null && groupedSections.length > 0) {
+      const direcao = groupedSections.find((s) => s.name === 'Direção' && s.accounts.length > 0);
+      if (direcao) {
+        openSection = direcao.name;
+      } else {
+        const firstNonEmpty = groupedSections.find((s) => s.accounts.length > 0);
+        openSection = firstNonEmpty?.name ?? null;
+      }
+    }
+  });
```
peucastro, line 60: The effect that initializes `openSection` searches for the 'Direção' string by its Portuguese name. This ties a runtime logic decision to a translated display string. If the display name ever changes (e.g. if we add more language options in the future), the priority will break. Consider giving the direction group a stable id field separate from its display name

peucastro (changes requested):
Overall this looks really good and is almost ready to merge! Just one thing I'd like to address before approving.

The roles translation and grouping logic in `groupedSections` is a bit hard to maintain right now. The role keys and their Portuguese translations appear as inline strings across the derived block, which makes it hard to follow.

It would be great to extract a small object at the top of the script, something like a `ROLE_DISPLAY_NAMES` record for translations, a `DIRECTION_ROLES` array for which roles get merged into "Direção", and an `EXCLUDED_ROLES` array for intentionally hidden roles (this is just a suggestion, feel free to implement it in whatever way you think it's best)

src/types/team-member.ts
```diff
@@ -5,7 +5,8 @@ export type TeamMember = {
   email: string;
   bio: string;
   birthDate: Date;
-  role: string;
+  role?: string;
```
peucastro: The `role` attribute is deprecated in the backend-side, and can be safely removed. Make sure to update the `team-member-hexagon.svelte` to remove the logic for checking for the `role` attribute.

src/routes/(app)/team/+page.svelte
```diff
@@ -0,0 +1,296 @@
...
+
+  const centerActiveSection = (animate: boolean = false) => {
+    if (!containerRef || orderedSections.length === 0) return;
+
+    const centerBtn = buttonRefs[CAROUSEL_CENTER_INDEX];
+    const prevBtn = buttonRefs[1];
+    const nextBtn = buttonRefs[3];
```
peucastro: It would be more consistent and clever to write `buttonRefs[CAROUSEL_CENTER_INDEX - 1]` and `buttonRefs[CAROUSEL_CENTER_INDEX + 1]` instead. That way, if the center index ever changes, the surrounding code stays correct.

src/routes/(app)/team/+page.svelte
```diff
@@ -0,0 +1,296 @@
...
+      indices = [2, 0, 1, 2, 3]; // [C, A, B, C, D]
+    } else {
+      indices = [4, 0, 1, 2, 3]; // no extra cycling
+    }
+
+    const carouselIds = ['buf-left', 'prev', 'curr', 'next', 'buf-right'];
+
+    return indices.map((idx, pos) => {
+      const section = orderedSections[idx];
+      const isDuplicate = indices.filter((i) => i === idx).length > 1;
```
peucastro: This will scan the entire array for every element, which is `O(n^2)`. However, since the array only has 5 elements, this is a minor issue, feel free to ignore.

peucastro (changes requested):

peucastro (approved):
LGTM! Thank you so much! ✅ 🚀

PR #388, opened 2025-12-20
Permissions updated

### LINKED ISSUE #384: auth: Centralize permission handling
Currently, permission checks are scattered across the codebase, with direct API calls (e.g., `/api/auth/hasPermission/:id`) and custom logic in multiple places.

We should create a single, reusable permissions utility or service and update all usages in the codebase to use the new centralized approach (already partially implemented in `permissions.ts`)

MiguelMoita:
Closes #384 

Changed the permissions to be used in a not hard coded and central way.


src/lib/api/permissions.ts
```diff
@@ -1,29 +1,48 @@
...
+  const perms = await fetch(`/api/auth/hasPermission/${number}`);
 
   if (!perms.ok) {
     return false;
   }
 
   return !(await perms.json()).error;
 }
+export async function canCreateAccount(fetch: typeof globalThis.fetch) {
+  return await hasPermission(fetch, 0);
```
peucastro: Instead of using magic numbers, it would be better to have a centralized `enum Permission` with a map to the permissions in the backend.

src/lib/api/permissions.ts
```diff
@@ -1,29 +1,48 @@
...
-    return false;
-  }
+export async function canEditAccount(fetch: typeof globalThis.fetch) {
+  return await hasPermission(fetch, 2);
+}
 
-  return !(await perms.json()).error;
+export async function canDeleteAccount(fetch: typeof globalThis.fetch) {
+  return await hasPermission(fetch, 3);
+}
```
peucastro, line 37: You can simplify these functions by removing the redundant `await` when directly returning a `Promise`. Since `hasPermission` already returns a `Promise`, you can just return it directly

peucastro (changes requested):

src/routes/(app)/projects/[slug]/+page.ts
```diff
@@ -1,15 +1,14 @@
...
 export const load: PageLoad = async ({ fetch, params }) => {
   let res = await fetch(`/api/projects/${params.slug}`);
   if (!res.ok) error(res.status, 'Project not found');
 
   const project = await res.json();
 
-  let hasPerms = false;
-  res = await fetch(`/api/auth/hasPermission/6`);
-  if (res.status !== 403) hasPerms = true;
+  let hasPerms = await canDeleteActivity(fetch);
```
Comment (tomaspalma) [inline src/routes/(app)/projects/[slug]/+page.ts]: This is a nitpick, but you can switch `let` to `const`, so it would become

```js
const hasPerms = await canDeleteActivity(fetch);
```

src/routes/(app)/projects/+page.ts
```diff
@@ -13,9 +14,7 @@ export const load: PageLoad = async ({ fetch }) => {
 
   const projects = await res.json();
 
-  let hasPerms = false;
-  res = await fetch(`/api/auth/hasPermission/4`);
-  if (res.status !== 403) hasPerms = true;
+  let hasPerms = await canCreateActivity(fetch);
```
tomaspalma: This is a nitpick, but you can switch `let` to `const`, so it would become

```js
const hasPerms = await canDeleteActivity(fetch);
```

tomaspalma (changes requested):

peucastro (approved):
Thank you! ✅ 🚀

PR #389, opened 2026-01-07
Fix/admin role page base

### LINKED ISSUE #386: admin: roles sub-page base layout

carolinaroquee:
Closes #386 

Fix Admin Roles Base Page

# Screenshots
Captura de ecrã 2026-01-07 154638

# Review checklist

- [ ] Behavior is as expected


src/routes/(app)/admin/_components/roles.svelte
```diff
@@ -0,0 +1,127 @@
...
+
+<section class="flex h-full w-full flex-row py-12">
+  <aside class="flex w-80 flex-col gap-y-4 pr-12">
+    {#each roles as role (role.id)}
+      <button
+        class="rounded-full px-8 py-4 text-left text-xl font-medium transition-all duration-200
+               {selectedRole?.id === role.id
+                 ? 'bg-red-600 text-white shadow-lg'
+                 : 'bg-white/10 text-white hover:bg-white/20'}"
+        on:click={() => (selectedRole = role)}
```
peucastro: Beware that in Svelte 5 the event handled syntax changed from `on:click` to `onclick` (more details [here](https://svelte.dev/docs/svelte/svelte-events))

src/routes/(app)/admin/_components/roles.svelte
```diff
@@ -0,0 +1,127 @@
...
+                    errorMessage = 'Role já existente!';
+                    return;
+                    }
+                    if (result.type === 'success' && result.data?.success) {
+                        const newRole = result.data.data as Role;
+                        roles = [...roles, newRole];
+                        selectedRole = newRole;
+                        dialogOpen = false;
+                    } else {
+                        errorMessage = result.data?.error || 'Erro ao criar a role';
```
peucastro: `result.data` may not exist on all action result types. Thus, this results in a type error

peucastro (changes requested):
Overall this looks pretty good and works as expected!

I just have a little nitpick to point out: the roles buttons are currently looking a bit off when comparing to the version in the mockups. Just Fix this and the other comments I've sent.

| Mockups      | Implementation      |
| ------------- | ------------- |
| image | image |

carolinaroquee:
image

Fixed the roles buttons and also handled the case where `result.data` may not exist on all action result types by checking `result.type` before accessing it.

peucastro (commented):
Amazing work! This PR is just missing responsiveness to be merged, but I've just noticed that we don't have mockups for the mobile version of this page 😕. We can gather some opinions anyway in today's meeting to see what we can do

src/types/role.ts
```diff
@@ -0,0 +1,7 @@
+export type Role = {
+  name: string;
+  permissions: string[];
+  isSection: boolean;
+  id: number;
+  associatedActivities: unknown[];
```
peucastro: If the API has a defined shape for these , it should be typed

src/routes/(app)/admin/+page.server.ts
```diff
@@ -20,7 +21,18 @@ export const load: PageServerLoad = async ({ fetch }) => {
...
   const technologies: Technology[] = await tech.json();
 
-  return { technologies };
+  const role = await fetch('/api/roles');
+  if (!role.ok) {
+    if (role.status === 404) {
+      return { roles: [] };
+    } else {
+      error(role.status, 'Failed to load roles');
+    }
```
peucastro: Ensure all return paths include both technologies and roles. If one fails with a 404, return an empty array for it so data.roles or data.technologies never becomes undefined in the UI.

src/routes/(app)/admin/+page.server.ts
```diff
@@ -79,5 +91,40 @@ export const actions: Actions = {
...
+    if (!(await canCreateActivity(fetch))) {
+      throw redirect(303, '/');
+    }
+
+    const formData = await request.formData();
+    const name = formData.get('name');
+
+    if (!name || typeof name !== 'string' || name.trim() === '') {
+      return { success: false, error: 'O nome da role é obrigatório' };
+    }
```
peucastro, line 105: Here we can use `error(400, ...)` to maintain consistency with the other with the other methods in this file

src/routes/(app)/admin/_components/roles.svelte
```diff
@@ -0,0 +1,171 @@
...
+  import Icons from '$lib/components/icons/icons';
+  import * as Dialog from '$lib/components/ui/dialog/index.js';
+  import LabelInput from '$lib/components/forms/label-input.svelte';
+  import { enhance, applyAction } from '$app/forms';
+  import { sentenceFirstLetterToUpperCase } from '$lib/utils';
+
+  type BackendError = {
+    message: string;
+    params?: Record<string, string>;
+  };
```
peucastro: This type is already defined in `src/types/backend-error.ts`

src/routes/(app)/admin/_components/roles.svelte
```diff
@@ -0,0 +1,171 @@
...
+                  id="name"
+                  name="name"
+                  type="text"
+                  required={true}
+                  class="rounded-lg bg-white px-4 py-3 text-black"
+                />
+
+                {#if errorMessage}
+                  <p class="text-center font-medium text-red-300">
+                    {sentenceFirstLetterToUpperCase(errorMessage)}
```
peucastro:
```suggestion
                    {sentenceFirstLetterToUpperCase(errorMessage.message)}
```

src/routes/(app)/admin/_components/roles.svelte
```diff
@@ -0,0 +1,171 @@
...
+  import * as Dialog from '$lib/components/ui/dialog/index.js';
+  import LabelInput from '$lib/components/forms/label-input.svelte';
+  import { enhance, applyAction } from '$app/forms';
+  import { sentenceFirstLetterToUpperCase } from '$lib/utils';
+
+  type BackendError = {
+    message: string;
+    params?: Record<string, string>;
+  };
+  let { roles = $bindable([] as Role[]) } = $props();
```
peucastro: Instead of using a type cast in the `$bindable` rune, the standard followed is to create an `Props` interface with the desired type, for example:

```ts
interface Props {
  roles: Role[];
}
```

src/routes/(app)/admin/_components/roles.svelte
```diff
@@ -0,0 +1,171 @@
...
+              } else if (result.type === 'failure') {
+                errorMessage = {
+                  message:
+                    typeof result.data?.error === 'string' ? result.data.error : 'Erro de validação'
+                };
+              } else {
+                errorMessage = { message: 'Erro inesperado' };
+              }
+
+              await applyAction(result);
```
peucastro: In `technologies.svelte`, `applyAction` is only called on the success path. Here it's called regardless of the result type, including the failure/error branches. It's inconsistent with the existing codebase. Consider calling `applyAction` only when `result.data?.success` is true (or `result.data.errors` is false).

peucastro (changes requested):

peucastro (approved):
Thank you so much! Amazing changes 🚀 ✅

PR #393, opened 2026-01-22
hide button

### LINKED ISSUE #376: User Page Edit Button Visible Without Ownership Check
Right now, the edit button is shown on `user/[id]` pages unconditionally.

It should only appear when the viewer is the owner of that profile or has edit permission.

yili717:
Closes #376

[Description of the changes proposed in the pull request]

# Screenshots

# Review checklist

- [ ] Behavior is as expected


src/routes/(app)/user/[id]/+page.ts
```diff
@@ -7,5 +7,21 @@ export const load: PageLoad = async ({ fetch, params }) => {
...
+  let canEdit = false;
+  let viewer = null;
+
+  const authRes = await fetch('/api/auth');
+  if (authRes.ok) {
+    const auth = await authRes.json();
+    viewer = auth?.authenticated_user;
+  }
+
+  if (viewer && String(viewer.id) === params.id) {
```
Comment (peucastro) [inline src/routes/(app)/user/[id]/+page.ts]: Currently, you're converting the viewer's ID to a string to compare with `params.id`. This works, but it's safer to convert `params.id` to a number instead, since IDs are numbers in the database:


```suggestion
  if (viewer && viewer.id === Number(params.id)) {
```

src/routes/(app)/user/[id]/+page.ts
```diff
@@ -7,5 +7,21 @@ export const load: PageLoad = async ({ fetch, params }) => {
...
+    canEdit = true;
+    // } else if (viewer) {
+    //   const permRes = await fetch('/api/auth/hasPermission/SUPERUSER');
+    //   if (permRes.ok) canEdit= true;
+  }
```
Comment (peucastro) [inline src/routes/(app)/user/[id]/+page.ts]: Right now the code for allowing superusers to edit any profile is commented out. This should be enabled, so admins can help users edit their profiles if needed. The logic looks correct, just needs to be uncommented and fixed:


```suggestion
if (viewer && viewer.id === Number(params.id)) {
  canEdit = true;
} else if (viewer) {
  const permRes = await fetch('/api/auth/hasPermission/9'); // 9 is SUPERUSER
  if (permRes.ok) {
    const result = await permRes.json();
    if (!result.error) canEdit = true;
  }
}
```

Notice that you need to check `!result.error` in the response body, not just if the fetch was successful.

peucastro (changes requested):

peucastro (approved):
Thanks! ✅ 💯

PR #413, opened 2026-03-25
Password recovery page

### LINKED ISSUE #394: auth: Implement the recover password page
Refer to the backend PRs in the `feature/create-account` (https://github.com/NIAEFEUP/website-niaefeup-backend/pull/260) and `feature/improved-password-recovery` (https://github.com/NIAEFEUP/website-niaefeup-backend/pull/264) branches for details on the new password recovery/creation flow.

When a user creates an account or requests to recover their password, they will receive an email with a link containing a signed token.

When the user opens this link, display a form with two fields: "New Password" and "Confirm Password". Validate that both fields match and that the password meets the required criteria. After the user submits the form, send the new password along with the token to the backend endpoint to set or update the password.

If the password is successfully changed, show a success message and redirect the user to the login page. If the link is expired or invalid, display an error message and prompt the user to request a new recovery email or account setup link.

This page should work for both account creation (setting the initial password) and password recovery, as both use the same backend endpoint.

David-Ferreira27:
Closes #394

Page dedicated to letting the user reset their password.
Two input fields for typing the new password and confirming the new password.
Password won't be changed if the new password is outside size boundaries or if it is different from the password typed in the confirm password field.

# Screenshots
image

image

# Review checklist

- [ ] Behavior is as expected

### DIFF
--- a/src/routes/(app)/auth/password/recovery/[token]/+page.server.ts
+++ b/src/routes/(app)/auth/password/recovery/[token]/+page.server.ts
@@ -1,8 +1,6 @@
 import type { PageServerLoad, Actions } from './$types';
 
 function getResult(pw: string, pw_confirm: string): string {
-    if (pw.length <= 7)
-        return "New password too short!";
     if (pw !== pw_confirm)
         return "Passwords don't match!";
     return "";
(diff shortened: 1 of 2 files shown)


peucastro (changes requested):
This works as expected, but there are some things that need to be addressed before we merge this:

- Use the existing forms components (`src/lib/components/forms/`), they provide more consistent styling, labels, and HTML validation attributes. Use it instead of the raw <input> elements to match the rest of the website.
- For error feedback, i would follow the contacts page pattern: call `createNotification()` from `src/routes/(app)/_components/layout/notifications/index.ts` on error, but you can research and debate better ways of doing this.
- Add a loading state to the submit button (see `src/routes/(app)/contacts/+page.svelte`) and disable it during submission to prevent double submits.
- Remove the console.log at `+page.server.ts` and use `autocomplete="new-password"` on the fields for better accessibility.

